PR #29, opened 2016-03-18
Add L3 binding support

### LINKED ISSUE #28: Add L3 binding support
Support L3 binding, including loading L3 binding class defined in configuration file, registering BIG-IP MAC addresses, and binding address/subnet pairs when load balancer is created.

jlongstaf:
Issues:
Fixes #28

Problem: Need to add L3 binding support`

Analysis: Added L3 binding calls in icontrol driver, and plugin rpc method.

Tests:


richbrowne:
Adding implementation of L3 Binding.

PR #53, opened 2016-03-30
Add source character encoding definition.

jlongstaf:
@jputrino @richbrowne 
#### What's this change do?
Adds character encoding definition to top of files: # coding=utf-8

Any file.

#### Any background context?
Previous change to add F5 trademarks included non-ASCII characters which causes imports to fail.


jputrino:
Looks good to me! I noticed that the copyright blurb is missing from one of the modules (lbaasv2/drivers/bigip/icontrol_driver.py); I created a new issue for it ( #54 ) and should be able to get that fix in today.

jlongstaf:
Jodie - Thanks for catching the missing copyright.

PR #107, opened 2016-05-12
Fix system V init script to reuturn correct status.

### LINKED ISSUE #106: Fix System V startup to return correct status
The System V startup script does not return correct status.

#### Agent Version
2.0.1

#### Operating System
Ubuntu

#### OpenStack Release
Liberty

#### Description
When the user runs # service f5-oslbaasv2-agent status, there is an error in finding the process.

richbrowne:
@jlongstaf @mattgreene 
Fixes #106 
...

#### What's this change do?
Fixes service status for agent daemon

f5-oslbaasv2-init

#### Any background context?

Issues:
Fixes #106

Problem:
The initialization script does not return status for the
agent daemon.

Analysis:
Fix logic to find agent daemon in ps output.

Test:
Run service f5-oslbaasv2-agent status.

### DIFF
--- a/etc/neutron/services/f5/f5-openstack-agent.ini
+++ b/etc/neutron/services/f5/f5-openstack-agent.ini
@@ -425,8 +425,7 @@ f5_common_external_networks = True
 # for all mapped fixed_ips will be made referencing the ports id
 # and the ports device_id.
 #
-# l3_binding_static_mappings = 'subnet_a':[('port_a','device_a'),('port_b','device_b')], 'subnet_b':[('port_c','device_a'),('port_d','device_b')]
-#
+# l3_binding_static_mappings = {"subnet_b": [["port_c", "device_a"], ["port_d", "device_b"]], "subnet_a": [["port_a", "device_a"], ["port_b", "device_b"]]}
 #
 #
 ###############################################################################
@@ -454,7 +453,7 @@ f5_bigip_lbaas_device_driver = f5_openstack_agent.lbaasv2.drivers.bigip.icontrol
 # self IPs, you should specify them as a comma
 # separated list below. 
 #
-icontrol_hostname = 10.190.7.232
+icontrol_hostname = 10.190.7.116
 #
 # If you are using vCMP® with VLANs, you will need to configure
 # your vCMP® host addresses, in addition to the guests addresses.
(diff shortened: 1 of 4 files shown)


etc/neutron/services/f5/f5-openstack-agent.ini
```diff
@@ -425,8 +425,7 @@ f5_common_external_networks = True
 # for all mapped fixed_ips will be made referencing the ports id
 # and the ports device_id.
 #
-# l3_binding_static_mappings = 'subnet_a':[('port_a','device_a'),('port_b','device_b')], 'subnet_b':[('port_c','device_a'),('port_d','device_b')]
-#
+# l3_binding_static_mappings = {"subnet_b": [["port_c", "device_a"], ["port_d", "device_b"]], "subnet_a": [["port_a", "device_a"], ["port_b", "device_b"]]}
```
mattgreene: Is this change necessary?

etc/neutron/services/f5/f5-openstack-agent.ini
```diff
@@ -454,7 +453,7 @@ f5_bigip_lbaas_device_driver = f5_openstack_agent.lbaasv2.drivers.bigip.icontrol
 # self IPs, you should specify them as a comma
 # separated list below. 
 #
-icontrol_hostname = 10.190.7.232
+icontrol_hostname = 10.190.7.116
```
mattgreene: We should get in the habit of not pushing changes to this variable.

f5_openstack_agent/lbaasv2/drivers/bigip/l3_binding.py
```diff
@@ -115,17 +122,10 @@ def bind_address(self, subnet_id=None, ip_address=None):
...
-                            port_id=port_id,
-                            ip_address=ip_address
-                        )
-                    else:
-                        LOG.error(
-                            'No RPC to plugin available to add '
-                            'allowed address %s to port: %s.'
-                            % (ip_address, port_id)
-                        )
+                    self.driver.plugin_rpc.add_allowed_address(
```
mattgreene: It's not clear from the commit message why this change is necessary.  You've remove some error handling.  Do we need a try/except around this call- even if the old if.self.driver.plugin.rpc check is no longer necessary?

etc/neutron/services/f5/f5-openstack-agent.ini
richbrowne: I agree, this was unintentional.

etc/neutron/services/f5/f5-openstack-agent.ini
richbrowne: Yes, the other line is not parseable.  If we provide an example, it should work with the code.

PR #124, opened 2016-05-26
Implemented Disconnected Service Deployment

### LINKED ISSUE #109: Disconnected Service Deployment
# Introduction
This document is meant to provide a lightweight, high-level view of the “Disconnected Service Deployment” capability in the F5 OpenStack Agent, hereafter known as “agent”.  The business driver for this capability is a requirement to support “Hierarchical Port Binding”; as possible via the ML2 plugin driver architecture.

# Background
The agent assumes that a Neutron network will already exist when attempting to create Neutron LBaaS objects.  The service model constructed and sent to the agent, and subsequent RPC from the agent back to Neutron for discovery purposes will always have complete information.  Therefore, the agent is able to immediately create the necessary configuration on attached BIG-IP(s).

Deploying an OpenStack network architecture with hierarchical port-binding breaks current agent assumptions.  The key difference is that the network to which the BIG-IP should attach may not exist at the time Neutron LBaaS sends loadbalancer requests to the F5 plugin driver.   The network design provides an outward-facing segment ID to connect to the public/provider network.  This ‘top’ network can cascade through a series of one or more intermediate segments before reaching the ‘bottom’ network to which the BIG-IP will be attached.

# Proposal
The agent will be modified to create a “disconnected network” in order to support the “delayed activation” of loadbalancer objects.  This will become the default behavior and not modifiable by the operator.  For deployments using pre-existing Neutron networks, there may be a minor increase (on the order of seconds) in the time to activate BIG-IP configuration.  For deployments that create networks on-the-fly, the agent will poll for a maximum of 5 minutes waiting for the BIG-IP facing network to become available.  The agent does not attempt to create a Virtual Server until both a loadbalancer AND listener object have been created.  The listener object will sit in “PENDING_CREATE” state until either the network is created or the 5-minute timeout occurs.

### Disconnected Network
The agent will create a VLAN interface named “disconnected-vlan” within each tenant partition that is not connected to an external interface.  Therefore, the interface will not be capable of sending or receiving data traffic.  But it provides sufficient configuration to support creation of loadbalancer objects.

### Delayed Activation
Specific network fields in the Service Model will be allowed to point to, or contain, a non-existent value.  The agent will retain its current behavior that immediately translates loadbalancer objects to the appropriate BIG-IP configuration.  But the Virtual Server/Listener will be connected to the tenant’s disconnected network.  For this proposal, the segment ID (i.e. VXLAN tunnel ID) can be empty even though the Service Model should still contain the name of the network missing an ID.

## Deliverables

* Enhanced F5 OpenStack agent

### Assumption
1. Operator creates loadbalancer with the parent subnet_id
2. ML2 plugin driver creates the network hierarchy
3. Agent must detect a hierarchical network and traverse down to locate the child network
4. Agent polls for child subnet_id
5. Agent requests a neutron port (for the listener, a.k.a. VIP) on the child subnet_id
6. Agent creates the tenant interface on BIG-IP and binds the listener (Virtual Server address) 

### Sequence Diagram
image

mattgreene:
@richbrowne @zancas 

Issues:
Fixes #109

Problem:
Agent presumes that neutron network are fully constructed prior to creating loadbalancer objects.

Analysis:
Modify agent behavior to detect incomplete network deployment, via the semengation_id field, and place a virtual server on a "disconnectd network".  The agent will periodically poll and move virtual servers off of the disconnected network once the seg_id exists.

Test:
Traffic test passes from lbaasv2_driver
New agent functional tests pass

Note:
Deferring testing that emulates customer network environment due to lack of actionable data from customer as well as the upcoming "standlaone agent test environment" that will enable granular control over the service model sent to the agent.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
@@ -541,7 +541,7 @@ class iControlDriver(LBaaSBaseDriver):
         if self.conf.f5_ha_type == 'scalen' and \
                 self.cluster_manager.get_sync_status(bigip) == 'Standalone':
             raise f5ex.BigIPClusterInvalidHA(
-                'HA mode is pair and bigip %s in standalone mode'
+                'HA mode is scalen and bigip %s in standalone mode'
                 % hostname)
 
         if self.conf.f5_ha_type != 'standalone':
@@ -578,10 +578,10 @@ class iControlDriver(LBaaSBaseDriver):
 
         if self.conf.f5_ha_type != 'standalone':
             if self.conf.f5_sync_mode == 'autosync':
-                self.cluster_manager.enable_auto_sync(bigip, device_group_name)
+                self.cluster_manager.enable_auto_sync(device_group_name, bigip)
             else:
-                self.cluster_manager.disable_auto_sync(bigip,
-                                                       device_group_name)
+                self.cluster_manager.disable_auto_sync(device_group_name,
+                                                       bigip)
 
         # Turn off tunnel syncing... our VTEPs are local SelfIPs
         if self.system_helper.get_tunnel_sync(bigip) == 'enable':


f5_openstack_agent/lbaasv2/drivers/bigip/disconnected_service.py
```diff
@@ -0,0 +1,149 @@
...
+                listeners = service.get('listeners', None)
+                service.pop('listeners', None)
+                provisioning_status = \
+                    service['loadbalancer']['provisioning_status']
+                if not d.disconnected_service.is_service_connected(service):
+                    if id in self.start_time:
+                        if timeutils.is_older_than(
+                                self.start_time[id],
+                                d.conf.f5_network_segment_gross_timeout):
+                            LOG.error("failed to connect loadbalancer %s to "
```
swormke: Should this log message be worded so that we know we hit the timeout versus some other error conditions?

f5_openstack_agent/lbaasv2/drivers/bigip/disconnected_service.py
```diff
@@ -0,0 +1,149 @@
...
+                                      "a real network" % id)
+                            if (provisioning_status !=
+                                    plugin_const.PENDING_DELETE):
+                                d.plugin_rpc.update_loadbalancer_status(id)
```
swormke: Probably a stupid question, but what is the status you are updating to?  I see a similar cal below where you update to `lb_const.ONLINE`, but don't see a status in this call.

f5_openstack_agent/lbaasv2/drivers/bigip/disconnected_service.py
```diff
@@ -0,0 +1,149 @@
...
+                    else:
+                        self.start_time[id] = timeutils.utcnow()
+                        d.plugin_rpc.update_loadbalancer_status(
+                            id, provisioning_status, lb_const.OFFLINE)
+                    continue
+                # service is connected in neutron, move all listeners for this
+                # loadbalancer onto a real network
```
swormke: Is is possible for a loadbalancer to have multiple listeners and therefore have one of the ports make it all the way through the portbinding, but one or more of the others not be there?

f5_openstack_agent/lbaasv2/drivers/bigip/disconnected_service.py
```diff
@@ -0,0 +1,149 @@
...
+        connected = True
+        for bigip in bigips:
+            vs = bigip.ltm.virtuals.virtual
+            if vs.exists(name=virtual['name'], partition=virtual['partition']):
+                vs.load(name=virtual['name'], partition=virtual['partition'])
+                if (getattr(vs, 'vlansDisabled', False) or
+                        not getattr(vs, 'vlansEnabled', True)):
+                    # accommodate quirk of how big-ip returns virtual server
+                    # if vlans are disabled OR vlans are not enabled, then
+                    # we're connected
```
swormke: I was thinking about this code and the BIG-IP's behavior and I think it is safe to say that you will always only ever have one or the other parameters in the JSON that is returned by BIG-IP and the one that is returned will always be the one that is set to `True`. I am fine how you have it, but you can probably just do the first one half and be fine.

```python
if (vs.get('vlansDisabled', False):
    ...
```

f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
```diff
@@ -1098,10 +1118,15 @@ def _update_loadbalancer_status(self, loadbalancer):
 
         if (provisioning_status == plugin_const.PENDING_CREATE or
                 provisioning_status == plugin_const.PENDING_UPDATE):
+            operational_status = lb_const.ONLINE
+            if self.disconnected_service_polling.enabled:
+                # operational status will be set by the disconnected
+                # service polling thread if that mode is enabled
+                operational_status = lb_const.OFFLINE
```
swormke: I guess this is why the call above works because you set it to OFFLINE by default?

f5_openstack_agent/lbaasv2/drivers/bigip/disconnected_service.py
mattgreene: The parameter defaults are ERROR and OFFLINE.

f5_openstack_agent/lbaasv2/drivers/bigip/disconnected_service.py
mattgreene: Sure.  I will clarify the message.

f5_openstack_agent/lbaasv2/drivers/bigip/disconnected_service.py
mattgreene: Yes, a loadbalancer can possess multiple listeners.  But the loadbalancer object contains the (one) subnet_id/network_id, on which all listeners sit.  So the agent needs to wait for that single vip network to become attached, but then migrate all listener children.

PR #128, opened 2016-05-27
Replace tmp file cert/key upload with memory buffer upload.

### LINKED ISSUE #127: Replace tmp file cert/key upload with memory buffer upload.
The initial implementation of creating SSL profiles for TERMINATED_HTTPS support temporarily wrote the cert/keys obtained from Barbican cert manager to the local controller file system. A new f5-sdk method was made available to prevent having to create temporary files and keep the cert/key data completely in memory. This issue is for replacing the temp file upload process with the byte buffer upload.

jlongstaf:
@<reviewer_id>
Fixes #127 

ssl_profile.py

#### Any background context?
New method to pass in bytes instead of file name was added to f5-sdk. Replaced
creating tmp file with byte method to keep it in-memory.

Issues:
Fixes #127

Problem: Replace creating a tmp file for cert/key upload with
string buffer.

Analysis: Used new interface in f5-sdk to keep cert/key data
in memory.

Tests: ssl_profile test.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/ssl_profile.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/ssl_profile.py
@@ -14,14 +14,13 @@
 # limitations under the License.
 #
 
-import os
-import shutil
-import tempfile
-
 from oslo_log import log as logging
-from requests import HTTPError
 
 LOG = logging.getLogger(__name__)
+
+
+class SSLProfileError(Exception):
+    pass
 
 
 class SSLProfileHelper(object):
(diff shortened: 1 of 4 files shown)


richbrowne:
See if there is a way to clear out the cert and keyfile memory buffers.  Not really sure how this might be done with python object.

f5_openstack_agent/lbaasv2/drivers/bigip/ssl_profile.py
```diff
@@ -39,42 +38,33 @@ def create_client_ssl_profile(bigip, name, cert, key, sni_default=False):
...
+            # In-memory upload -- data not written to local file system but
+            # is saved as a file on the BIG-IP.
+            uploader.upload_bytes(cert, certfilename)
+            uploader.upload_bytes(key, keyfilename)
+
+            # import certificate
+            param_set = {}
+            param_set['name'] = certfilename
+            param_set['from-local-file'] = \
+                '/var/config/rest/downloads/' + certfilename
```
swormke: probably want to use the os.path.join for this type of thing.

https://docs.python.org/2/library/os.path.html

jlongstaf:
@richbrowne: After investigating erasing memory it seems that there is not much we can do. The bottom line is that strings in Python are immutable and short of custom C code, the best we can do is to be aggressive in ensuring that the cert/key variables are deleted. See http://stackoverflow.com/questions/728164/securely-erasing-password-in-memory-python for a good summary of the problem.

I added  try block and del of cert and key to ensure that the memory is marked for garbage collection. Also, I added @swormke suggestion to used join() for the path.

PR #131, opened 2016-06-01
Agent fails to start with f5_ha_type not set to standalone

### LINKED ISSUE #130: Agent fails to start with f5_ha_type not set to standalone.
Setting the agent ini file with f5_ha_type set to pair or scalen causes an error because iControl driver code attempts to enable/disable auto sync. Agent starts fine if f5_ha_type is set to standalone.

jlongstaf:
@mattgreene 
Fixes #130

#### What's this change do?
Two things:
  1) fixes passing positional arguments to enable/disable_auto_sync()
in the correct order, and
  2) fixes an incorrect exeption message.
  Compare enable/disable_auto_sync change with method signature defined in cluster_manager.py.

#### Any background context?
No.

Issues:
Fixes #130

Problem: Agent fails to start when f5_ha_type is set to pair
or scalen (not standalone).

Analysis: When f5_ha_type is not standalone, the agent either
enables or disables auto sync. icontrol driver code was passing
positional arguments to cluster_manager.enable_auto_sync in
wrong order. Fixed to pass arguments in correct order.

Tests: Manual


richbrowne:
Reviewed with Jeff.

PR #134, opened 2016-06-02
Feature.disconnected service

### LINKED ISSUE #109: Disconnected Service Deployment

mattgreene:
@richbrowne @zancas 

Issues:
Fixes #109

Problem:
Agent presumes that neutron network are fully constructed prior to creating loadbalancer objects.

Analysis:
Modify agent behavior to detect incomplete network deployment, via the semengation_id field, and place a virtual server on a "disconnected network".  The agent will periodically poll and move virtual servers off of the disconnected network once the semengation_id exists.

Test:
Traffic test passes from lbaasv2_driver
New agent functional tests pass

Note:
Deferring testing that emulates customer network environment due to lack of actionable data from customer as well as the upcoming "standalone agent test environment" that will enable granular control over the service model sent to the agent.


mattgreene:
Creating new pull request for feature.disconnected-service implementation into an isolated branch from liberty to wait for customer feedback on solution.

PR #160, opened 2016-06-15
Virtual Server should not listen on all tunnels/vlans

### LINKED ISSUE #137: Virtual Server should not listen on all tunnels/vlans
#### Agent Version
8.0.1

#### Operating System
N/A

#### OpenStack Release
Liberty

#### Description
User creates a loadbalancer and listener via neutron cli.  The F5 agent translates the listener definition into a virtual server definition that is sent to the BIG-IP specified in the agent .ini file.  The virtual server is not constrained a specific set of VLANs meaning that it will "live" on all networks accessible by the tenant.  The virtual server _should_ always have vlansEnabled = True and the vlans list constrained to the specific vlan or tunnel networks constructed for the tenant.

#### Deployment
N/A

richbrowne:
@jlongstaf @mattgreene 
Fixes #137 

#### What's this change do?
Pins Virtual Server to the specific VLAN/Tunnel of the VIP.

#### Any background context?

Issues:
Fixes #137

Problem:
The virtual server is not constrained a specific set of VLANs meaning
that it will "live" on all networks accessible by the tenant. The virtual
server should always have vlansEnabled = True and the vlans list constrained
to the specific vlan or tunnel networks constructed for the tenant.

Analysis:
Convert bigip.assured_networks to a dictionary where the key is
the OpenStack network_id and the value is the network name as it
appears in the BigIP configuration.

Save the names of the vlans/tunnels in the assured_networks.  When
a listener is created refer to the assured_networks map to extract
the vlan name of the virtual server's vip.

Tests:
Traffic/2-armed deployment.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/listener_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/listener_service.py
@@ -56,7 +56,9 @@ class ListenerServiceBuilder(object):
             tls['name'] = vip['name']
             tls['partition'] = vip['partition']
 
+        network_id = service['loadbalancer']['network_id']
         for bigip in bigips:
+            self.service_adapter.get_vlan(vip, bigip, network_id)
             self.vs_helper.create(bigip, vip)
 
             if tls:
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
@@ -582,7 +582,7 @@ class iControlDriver(LBaaSBaseDriver):
                   (bigip.device_name, ', '.join(bigip.mac_addresses)))
         bigip.device_interfaces = \
             self.system_helper.get_interface_macaddresses_dict(bigip)
-        bigip.assured_networks = []
+        bigip.assured_networks = {}
         bigip.assured_tenant_snat_subnets = {}
         bigip.assured_gateway_subnets = []
 
@@ -691,7 +691,7 @@ class iControlDriver(LBaaSBaseDriver):
     def flush_cache(self):
         # Remove cached objects so they can be created if necessary
         for bigip in self.get_all_bigips():
-            bigip.assured_networks = []
+            bigip.assured_networks = {}
             bigip.assured_tenant_snat_subnets = {}
             bigip.assured_gateway_subnets = []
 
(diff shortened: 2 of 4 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/listener_service.py
```diff
@@ -56,7 +56,9 @@ def create_listener(self, service, bigips):
             tls['name'] = vip['name']
             tls['partition'] = vip['partition']
 
+        network_id = service['loadbalancer']['network_id']
         for bigip in bigips:
+            self.service_adapter.get_vlan(vip, bigip, network_id)
```
mattgreene: This diff only shows new code to explicitly add vlan/tunnel ID to the allowed list for a VS.  How do vlan/tunnel IDs get added for members connected a different network than the listener?  I'd expect to see a similar get_vlan() call for each network ID in service['loadbalancer'['networks'].

richbrowne: These tunnels get created as part of prep_service_networking in the call to assure_bigip_network().  The call to get_subnets_to_assure() collects a list of subnets in a service, including the subnets for members, and iterates over that list to create or validate the existence of tunnels/vlans for each.  This call is for populating the vlan list in the virtual server only so maybe this should be named differently.

mattgreene: Let me ask the question differently.  The VS is the aggregation point for listeners and members.  This PR locks down the VLAN or tunnel added the VS to restrict traffic on the listener side.  But do we also need to add VLAN or tunnel to the VS for members that live on other networks than the listeners?

richbrowne: I am not sure about that.  Since the members' responses are routed back to the BigIP via the SNAT address, it isn't necessary, but I can't speak to why it would be desirable or not.  It might be a question that should be referred to John.  It should be easy to extend the functionality to add member networks, but my thought would be that the code could be implemented in the pool service.

Rich

mattgreene: Alright.  I'm ok with this for now provided the traffic test passes for both 1-armed and 2-armed deployments.  Did you already run both variations?

richbrowne: I believe I only ran the two armed variant.  My test environment was set up for that and I probably didn't run the 1 armed test.

mattgreene: Ok.  2-armed is the one likely to encounter a problem (with members sitting on 'other' neutron networks).

PR #168, opened 2016-06-17
Member static ARP population not performed.

### LINKED ISSUE #133: Member arp population not performed with FDB_POPULATE_STATIC_ARP == true
I noticed an error on cleanup of members for the LBaaSv2 service that indicated that the static arp cleanup could not be performed.

2016-06-02 11:58:18.445 57667 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.network_helper [req-bc143fc9-8a4a-476e-86f7-545074b3fd07 32da2ba1fb00494cb4b9d05528fa411a eca0dd84ec714e27a5310b9
bc5f30a36 - - -] Error deleting arp 10.2.2.2%251. Repsponse status code: 500. Response message: 500 Unexpected Error: Internal Server Error for uri: https://10.190.7.173:443/mgmt/tm/net/arp/
~Project_eca0dd84ec714e27a5310b9bc5f30a36~10.2.2.2%251

We are not actually populating this when the member gets created.  Though it looks like it should be there.  I looked at the LBaaSv1 code and did not see where this was added when a member is created.

richbrowne:
@jlongstaf @mattgreene 

Fixes #133
#### What's this change do?

Fixes the implementation of ARP lifecycle management in the icontrol driver; however, there are some issues that need to be resolved in either the BigIP REST implementation or the F5 SDK.

We have had a long standing error in the logs where there is a 500 error reported when we attempt to delete a static ARP entry that should be present.  It is not there, but it should and we fail to delete it resulting in an error.  However, instead of getting a 404 error, we see a 500 error when we test for existence.  The other problem is that the REST api reports HTTP 200 when we delete elements of a collection.
See f5-sdk issues:
https://github.com/F5Networks/f5-common-python/issues/495

network-helper.
#### Any background context?

Issues:
Fixes #133

Problem:
An error in the BigIP ARP implementation is hampering ARP
population management in partitions other than Common.  There
are a couple of issues:
1) An error in the implementation prevents the creation of a
static ARP entry for members in a pool.
2) An error in the BigIP REST API results in a HTTP 500 error
when an ARP entry in a partion other than Common is queried for
the existence of the ARP entry.
3) The BigIP REST API returns HTTP 200 OK when a ARP entry is
deleted even though the ARP still exist in the network config.

Analysis:
This commit fixes the error described in item 1 and allows us
to create the ARP; however, we still cannot query/delete the ARP
entry due to issues in the REST API.  This requires a fix
in the icontrol_rest api or an alternate SDK implementation
that relies on SOAP.

This commit also sets the default setting of f5_static_arp_population
from True to False.

Tests:
Traffic tests.

### DIFF
--- a/etc/neutron/services/f5/f5-openstack-agent.ini
+++ b/etc/neutron/services/f5/f5-openstack-agent.ini
@@ -213,7 +213,7 @@ f5_vtep_selfip_name = 'vtep'
 # record being added, that a static arp entry will be created to 
 # avoid the need to learn the member's MAC address via flooding.
 #
-# f5_populate_static_arp = True
+# f5_populate_static_arp = False
 #
 # Device Tunneling (VTEP) self IPs
 #
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
@@ -357,17 +357,15 @@ class NetworkHelper(object):
                    ip_address,
                    partition=const.DEFAULT_PARTITION):
         if ip_address:
-            address = urllib.quote(self._remove_route_domain_zero(ip_address))
-            arp = bigip.tm.net.arps.arp
             try:
-                if arp.exists(name=address, partition=partition):
-
-                    arp.load(name=address, partition=partition)
-                    arp.delete()
+                arps = bigip.tm.net.arps.get_collection(partition=partition)
+                for arp in arps:
+                    if arp.ipAddress == ip_address:
+                        arp.delete()
             except HTTPError as err:
                 LOG.error("Error deleting arp %s. "
                           "Repsponse status code: %s. Response "
-                          "message: %s." % (address,
+                          "message: %s." % (ip_address,
                                             err.response.status_code,
                                             err.message))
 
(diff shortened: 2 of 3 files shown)


mattgreene:
Please file a github issue against f5-sdk and link here to track the residual issue(s) that you mention in the first section.

mattgreene:
Did you run traffic tests for both under and overcloud?

etc/neutron/services/f5/f5-openstack-agent.ini
```diff
@@ -213,7 +213,7 @@ f5_vtep_selfip_name = 'vtep'
 # record being added, that a static arp entry will be created to 
 # avoid the need to learn the member's MAC address via flooding.
 #
-# f5_populate_static_arp = True
+# f5_populate_static_arp = False
```
mattgreene: Seems like doc and QE teams need to know this change is coming.
- documentation describes each knob
- tests might confirm default settings or now need to explicitly change a setting to preserve the old behavior

f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
```diff
@@ -357,17 +357,15 @@ def arp_delete(self,
...
-            arp = bigip.tm.net.arps.arp
             try:
-                if arp.exists(name=address, partition=partition):
-
-                    arp.load(name=address, partition=partition)
-                    arp.delete()
+                arps = bigip.tm.net.arps.get_collection(partition=partition)
+                for arp in arps:
+                    if arp.ipAddress == ip_address:
+                        arp.delete()
```
mattgreene: The GUI allows me to create a static arp entry with '%0' for the default domain but strips it out automatically.  It keeps anything else, like '%1'.

It looks like arp_delete might be called with an ip_address always containing '%<num', even for default domain.  That means the for loop won't delete arp entries for the common partition such as for shared network across tenants.

richbrowne:
Just undercloud.  The code is never executed in the other case.

richbrowne:
Matt I filed github issue number #495 in f5-common-python before I left.  Static arp population cannot be used without it.  I just fixed the code review with a link to this issue.

I changed the option because 1) static arp population is currently completely broken so changing it to false only removes the error we see in the log about not being able to remove an ARP entry and 2) with the changes that I have made we are now able to at least create an ARP entry, but there is no way to remove it without a fix in the f5-sdk or BigIP REST.  So setting it to true with these changes would make matters worse b/c there would be no way to clean up  a folder.

If this doesn't get fixed in this upcoming release by f5-sdk, then maybe we document the known issue in release notes, like static_arp_population is not supported.  So ARP discovery needs to be performed and the ARP gets put in dynamic arp list.

You found a bug in the last line.  I was trying to work around the limitations of the REST API, but that didn't work.  This code in delete should revert to the original version.

Thanks.

jgruber:
I need to test any ramifications of placing all static ARP entries in the /Common partition regardless of tenant. The IP addresses will be decorated with the route domain still. 

I agree we need to make the default behavior of the agent not to add the static entries. Dynamic learning still works, and we only need this when the environment gets large.

This is a blocker for migrating off of SOAP and the associated set_active_folder problems which are causing:

This is a blocker for migrating off of SOAP and the associated set_active_folder problems which are causing:

[https://github.com/F5Networks/f5-openstack-lbaasv1/issues/125](url)

mattgreene:
This agent bug is not a blocker for lbaasv1, which bundles in its own equivalent of the agent code.  The same fix _may_ need to be applied to both lbaasv1 and agent repos.  But we're not going to refactor lbaasv1 to use this agent.

swormke:
@richbrowne @mattgreene this is a very stale PR, what do you guys want to do with it? Leave it on a branch in someone's repo and close or is this something that actually needs to be addressed and completed?

richbrowne:
Needs to be addressed and completed.  There was some work needed in the SDK to fix this.

Rich

PR #179, opened 2016-07-07
f5-sdk requires assigning return values for load() and create()

### LINKED ISSUE #178: f5-sdk requires assigning return values for load() and create()
Usage of f5-sdk load() and create() methods is inconsistent in f5--openstack-agent code. With v1.0.0 of f5-sdk return values from load() and create() must be assigned to correctly use the object. Some parts of the agent code already assign return values, but some do not. Need to fix all code to use return values.

jlongstaf:
@mattgreene 
Fixes #178 

#### What's this change do?
Adds assignment of return values for f5-sdk load() and create() methods.

Anywhere.

#### Any background context?
f5-sdk originally updated the object that invoked create() and load() methods, but now returns updated objects. Callers must assign the return value if they need to subsequently use the object.

Issues:
Fixes #178

Problem: Changes to f5-sdk make it necessary to always assign
the return value from load() and create().

Analysis: Added return value assignments.

Tests: test_solution.py


zancas:
Let's remove `test___common_service_handler.py` it's not useful.

PR #197, opened 2016-07-15
Issue #194 Agent filter syntax is not compatible with REST API in ver…

### LINKED ISSUE #194: Agent filter syntax is not compatible with REST API in version 11.5.x
#### OpenStack Release
Liberty

#### Description
In BIG-IP version 11.5.4 the syntax used in the agent code is not compatible with the BIG-IP REST API because the `$` is encoded to `%24` and the REST server does not decode it back into a `$`.

This [DevCentral article](https://devcentral.f5.com/questions/icontrol-rest-filter-processing-not-http-compliant) explains the issue and the workaround.

#### Agent Version
8.0.4

#### Operating System
CentOS 7

#### OpenStack Release
Liberty, Mitaka

#### Description
Adding or removing configuration will cause this bug to happen when running against a 11.5.x image.

#### Deployment
Any deployment using BIG-IP 11.5.x will have this issue with the 8.0.2 agent configuration.

swormke:
@mattgreene 
Issue #194 Agent filter syntax is not compatible with REST API in version 11.5.x

Issues:
Fixes #194

Problem:
In BIG-IP version 11.5.4 the syntax used in the agent code is not
compatible with the BIG-IP REST API because the $ is encoded to
%24 and the REST server does not decode it back into a $.

This DevCentral article explains the issue and the workaround.
https://devcentral.f5.com/questions/icontrol-rest-filter-processing-not-http-compliant

Analysis:
* Created a get_filter function to get the right syntax based on the version of BIG-IP
* Used that function to construct the requests_params for the filter.

Tests:

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
@@ -18,6 +18,7 @@ import netaddr
 import os
 import urllib
 
+from f5_openstack_agent.lbaasv2.drivers.bigip.utils import get_filter
 from oslo_log import helpers as log_helpers
 from oslo_log import log as logging
 from requests.exceptions import HTTPError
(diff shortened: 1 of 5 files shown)


mattgreene:
Ran traffic tests against 11.5.4 and 11.6.0, and manually confirmed in agent log file that correct form of $filter is passed in URI.

PR #198, opened 2016-07-15
Issue #195 No obvious way to remove cert_manager support

### LINKED ISSUE #195: No obvious way to remove cert_manager support
#### OpenStack Release
Liberty or Mitaka

#### Description
If no cert_manager support is desired, there is no way to do this without causing a ERROR message in the log.

In the icontrol_driver.py module (where the oslo config setting is for cert_manager) we have this line:

```
self.cert_manager = None  # overrides register_OPTS
```

This does not 'override' the default value from self.conf. 

When we get to this code:

```
        if self.conf.cert_manager:
            try:
                self.cert_manager = importutils.import_object(
                    self.conf.cert_manager, self.conf)
            except ImportError:
                self.cert_manager = None
                LOG.error('Failed to import CertManager: %s'
                          % self.conf.cert_manager)

        if not self.cert_manager:
            LOG.debug('No CertManager is configured.')
```

if there is no cert_manager defined in the ini file, the default is taken and we try to load the Barbican certificate manager. 

If you don't want to load the Barbican certificate manager, but instead want to proceed without one, the proper thing to do is to create a NullCertificateProvider class which does not thing but debug logging. We can discuss if that should not be the default. Today if you set the cert_manager=None in the ini file we will try to import that as a python class and will throw an ERROR 'Failed to import CertManager: None'. 

As it is, if the customer does not have Barbican setup.. we dump errors.

#### Agent Version
9.0.1

#### Operating System
Ubuntu 14.04
RHEL 7.2

#### OpenStack Release
Mitaka

swormke:
@mattgreene @jgruber 

**This should not be merged until @mattgreene has tested it**
Issues:
Fixes #195

Problem:
If no cert_manager support is desired, there is no way to do this
without causing a ERROR message in the log.

Analysis:
* The default setting for cert_manager was set to the F5 cert manager
  driver instead of None.
* This caused it to always enter the section where it was imported
  and if the user does not have Barbican setup they will get an
  error in the log.

Tests:

### DIFF
--- a/etc/neutron/services/f5/f5-openstack-agent.ini
+++ b/etc/neutron/services/f5/f5-openstack-agent.ini
@@ -469,7 +469,7 @@ icontrol_password = admin
 ###############################################################################
 # Certificate Manager
 ###############################################################################
-cert_manager = f5_openstack_agent.lbaasv2.drivers.bigip.barbican_cert.BarbicanCertManager
+# cert_manager = f5_openstack_agent.lbaasv2.drivers.bigip.barbican_cert.BarbicanCertManager
 #
 # Two authentication modes are supported for BarbicanCertManager:
 #   keystone_v2, and keystone_v3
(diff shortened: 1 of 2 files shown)


mattgreene:
Tested on 11.6.0 with cert_manager ini setting commented (produces DEBUG message) and uncommented (produces original ERROR message)

PR #202, opened 2016-07-16
vlans fields in JSON when creating VirtualServer behave differently b…

### LINKED ISSUE #201: Unable to pass traffic through listener on BIG-IP 11.5.x
#### OpenStack Release
Liberty

#### Description
Client attempt to curl through BIG-IP fails with "No route to host".  Client can ping BIG-IP selfip, but unable to connect through the VIP attached to a Virtual Server.
 
#### Agent Version
8.0.4

#### Operating System
CentOS 7

#### OpenStack Release
Liberty, Mitaka

#### Description
Adding or removing configuration will cause this bug to happen when running against a 11.5.x image.

#### Deployment
Create LBaaSv2 loadbalancer with one listener, pool and member all attached to the same neutron network.  BIG-IP VE is deployed in the same tenant as client and server.

#### Analysis
In BIG-IP version 11.5.4 the REST server accepts vlansEnabled = False, but returns vlansEnabled = True in the response.  The solution that works for all releases of BIG-IP is to set vlansDisabled = True.

mattgreene:
@zancas 

Code fix identified by @swormke, tested and being submitted by me.

Issues:
Fixes #201

Problem:
In BIG-IP version 11.5.4 the REST server accepts vlansEnabled = False, but returns vlansEnabled = True in the response.

Analysis:
The solution that works for all releases of BIG-IP is to set vlansDisabled = True.

Tests:
Traffic test on 11.5.4 overcloud deployment (which is the only toplogy that fails for this reason).  Undercloud deployment is not susceptible to this problem.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
@@ -348,6 +348,7 @@ class ServiceModelAdapter(object):
             vip['vlans'].append(
                 bigip.assured_networks[network_id])
             vip['vlansEnabled'] = True
+            vip.pop('vlansDisabled', None)
 
     def _add_bigip_items(self, listener, vip):
         # following are needed to complete a create()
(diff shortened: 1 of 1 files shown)


zancas:
LGTM.

PR #229, opened 2016-07-26
Merge liberty into mitaka

mattgreene:
@jlongstaf 

Merge changes made in liberty since 8.0.4 forward into mitaka.


requirements.test.txt
```diff
@@ -0,0 +1,10 @@
+#Install the agent
+.
+# app
+f5-sdk
+git+https://github.com/openstack/neutron#stable/liberty
```
jlongstaf: The liberty reference will need to be replaced with mitaka after merging.

PR #248, opened 2016-08-02
Listener update with name change fails

### LINKED ISSUE #247: Listener update with name change fails
Updating a listener that includes a name change (e.g., lbaas-listener-update) fails because agent attempts to create a new virtual server with different name while old virtual server still exits.

BIGI-IP does not support renaming virtuals throught REST, but can implement a rename operation that deletes existing virtual and then creating a new virtual with new name.

jlongstaf:
@richbrowne 
Fixes #247 

#### What's this change do?
Adds ability to rename the virtual server associated with a listener object. Does so
by checking if an update includes a name change, and if it does, deletes existing
virtual and then creates a new one.

lbaas-builder.py, _assure_listeners_created()

#### Any background context?
Found when running tempest tests.


richbrowne:
Ship it.

PR #252, opened 2016-08-04
If cert manager is not loaded agent should show as down

### LINKED ISSUE #211: If cert manager is not loaded agent should show as down
* Title: A short but descriptive summary of the issue, whether it be a bug or enhancement.
* Attachments: For bugs, attach the agent log and configuration files
  * /etc/neutron//services/f5/f5-openstack-agent.ini
  * /var/log/neutron/f5-openstack-agent.log
* Details: For enhancements, fill out the template below.

#### OpenStack Release
Liberty

#### Description
If the user enables the certificate manager, but the manager cannot be loaded the agent should not start and the agent should show as down.

#### Agent Version
8.0.5

#### Operating System
HP HOS 3.0, but applies to all

#### OpenStack Release
Liberty

#### Deployment
This can happen for a few reasons such as:
1. User provided a certManager that could not be found or is not in the Python path
2. Certificate service cannot be reached or some other error on import of the manager.

jlongstaf:
@richbrowne @mattgreene 
Fixes #211 

#### What's this change do?
Adds additional checks for Python library import errors, as well as adding a retry loop in case Keystone or Barbican are in the process of starting. Modified icontrol driver raise exception in case of errors so that agent will exit.

Start with barbican_client.py then review icontrol_drvier.py

#### Any background context?
SAP requested retry loop as their services run in containers and may not be immediately available when agent starts.

Issues:
Fixes #211

Problem: Agent continues running if it fails to load barbican
client code.

Analysis: Added checks for import errors of Python libraries and
retry loop to ensure that client can talk with Barbican. Also,
added raising exception if error detected when initializing
barbican client code so that agent will not contiue.

Tests: Manual.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/barbican_cert.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/barbican_cert.py
@@ -14,13 +14,24 @@
 # limitations under the License.
 #
 
-from barbicanclient import client
+try:
+    from barbicanclient.client import Client
+except ImportError:
+    raise ImportError("Missing Python library barbicanclient."
+                      "Install barbicanclient and restart the agent.")
+try:
+    from keystoneauth1.identity import v2
+    from keystoneauth1.identity import v3
+    from keystoneauth1.session import Session
+except ImportError:
+    raise ImportError("Missing Python library keystoneauth1."
+                      "Install keystoneauth1 and restart the agent.")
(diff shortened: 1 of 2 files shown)


richbrowne:
Reviewed with Jeff and Matt.

richbrowne:
Reviewed With Jeff and Matt.

PR #258, opened 2016-08-05
Use keystoneclient as fallback for Barbican client auth

### LINKED ISSUE #257: Use keystoneclient  as fallback for Barbican client auth
The current implementation of BarbicanCertManger uses keystoneauth1 for establishing an authenticated session with Barbican. Some OpenStack installations may not have keystoneauth1 installed, so allow keystoneclient to be used if keystoneauth1 is not installed.

jlongstaf:
@mattgreene 
Fixes #257

#### What's this change do?
Uses keystoneclient modules for authenticating if keystoneauth1 is not installed.

barbican_cert.py

#### Any background context?
Using keystoneauth1 is preferred, but some installations may not have keysonteauth1 installed. Fix will allow installations that have keystoneauth1 to operate as before, but still allow systems that only have keystoneclient to function.

Issues:
Fixes #257

Problem: Original implementation uses keystoneauth1 to establish
an authenticated session for the barbican client. Some installations
may not have keystoneauth1 installed and the agent fails.

Analysis: Added using keystoneclient as a fallback. First try
to import keystoneauth1 modules, and if that fails, try to import
keystoneclient modules.

Tests: Manual.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/barbican_cert.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/barbican_cert.py
@@ -20,12 +20,22 @@ except ImportError:
     raise ImportError("Missing Python library barbicanclient."
                       "Install barbicanclient and restart the agent.")
 try:
+    # Try keystoneuath1 first as OpenStack is migrating from keystoneclient
+    # to keystoneauth1. Some systems, particularly liberty, may not
+    # have keystoneauth1 installed. If ImportError, try keystoneclient.
     from keystoneauth1.identity import v2
     from keystoneauth1.identity import v3
     from keystoneauth1.session import Session
+
 except ImportError:
-    raise ImportError("Missing Python library keystoneauth1."
-                      "Install keystoneauth1 and restart the agent.")
+    try:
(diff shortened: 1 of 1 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/barbican_cert.py
```diff
@@ -20,12 +20,22 @@
...
-    raise ImportError("Missing Python library keystoneauth1."
-                      "Install keystoneauth1 and restart the agent.")
+    try:
+        # no keystoneauth1 -- try keystoneclient
+        from keystoneclient.auth.identity import v2
+        from keystoneclient.auth.identity import v3
+        from keystoneclient.session import Session
+    except ImportError:
+        raise ImportError("Missing Python library keystoneclient."
+                          "Install keystoneclient and restart the agent.")
```
mattgreene: This import error message is fine for liberty.  But I'd like to see the original message returned for mitaka to install keystoneauth1.  Or in the merge from liberty to mitaka, we retain the original mitaka code (dummy commit).

PR #300, opened 2016-09-01
Route domain id allocation failure.

### LINKED ISSUE #299: route route domain allocation failure
When creating a new route domain the agent gets the set of all route domains on the BigIP, and
selects the lowest available route domain id available.  This allocation fails when there is a hole
in the list of existing route domains b/c the lowest available route domain id is not returned.

richbrowne:
@mattgreene 
Fixes #299 

#### Any background context?

Issues:
Fixes #299

Problem:
When creating a new route domain the agent gets the set of all route
domains on the BigIP, and selects the lowest available route domain id
available.

This allocation fails when there is a hole in the list of existing route
domains b/c the lowest available route domain id is not returned.

Analysis:
Return route domain id at the end of the function.

Tests:
Create a loadbalancer/listener on BIGIP system where a hole in the route domain
id list exists.
Reran Tempest TLS test suite.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
@@ -226,8 +226,8 @@ class NetworkHelper(object):
             else:
                 raise LookupError(
                     "The list of route domain ids is out of order")
-        else:
-            return lowest_available_index
+
+        return lowest_available_index
 
     @log_helpers.log_method_call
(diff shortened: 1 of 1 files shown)


mattgreene:
Here's why this code change is necessary:
http://python-notes.curiousefficiency.org/en/latest/python_concepts/break_else.html

The else clause on a for loop only executes when the loop exits without hitting a 'break'.  Good find.

PR #320, opened 2016-09-20
Feature.stats tests

### LINKED ISSUE #302: StatsHelper should cast the stats as integers and deal with 'nan'
#### OpenStack Release
Liberty+

#### Description
The stats that are returned by the `get_global_stats()` are strings and are blindly cast as integers through out the code base.  The problems is that in some circumstances these can be 'nan' and the `int()` call will throw.

We should cast these to ints in the get stats function so people don't need to worry about their types later.

#### Agent Version
8.0.7

### LINKED ISSUE #317: Unit tests for stats_helper.py
#### OpenStack Release
Liberty

#### Description
Add unit tests for the `stats_helper.py` functions.

### LINKED ISSUE #318: Remove unused score functions in the stats_helper.py file
#### OpenStack Release

Liberty+
#### Description

There are a number of methods in the `stats_helper` module that are not used by the OpenStack code.  They were left over from a previous version of the library that was used in a different project. This is now dead code and should be removed.

### LINKED ISSUE #319: Unit Tests for ssl_profile.py
#### OpenStack Release

Liberty
#### Description

Add unit tests for the ssl_profile.py functions.

swormke:
@jlongstaf @pjbreaux 
Fixes #319 
Fixes #318
Fixes #317 
Fixes #302  

#### What's this change do?
* Adds unit tests for two modules to get 100% line coverage
* Removes some dead code
* Fixes a bug if the output from the BIG-IP returns a non-integer for a stat like _nan_

* Look at the unit tests
* Look at the removed code, according to @jgruber it should not have been included and was from a previous use of this code.
* Look at the change to account for non digit stats.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_ssl_profile.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_ssl_profile.py
@@ -1,0 +1,119 @@
+# coding=utf-8
+# Copyright 2014-2016 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from f5_openstack_agent.lbaasv2.drivers.bigip.ssl_profile import \
+    SSLProfileError
+from f5_openstack_agent.lbaasv2.drivers.bigip.ssl_profile import \
+    SSLProfileHelper
+
+
+import mock
+import pytest
+
+
+class TestSSLProfileHelper(object):
+    def test_get_ssl_profile_count(self):
+        bigip = mock.MagicMock()
+        bigip.tm.ltm.profile.client_ssls.get_collection =\
+            mock.MagicMock(return_value=[1, 2, 3, 4])
+        assert(SSLProfileHelper.get_client_ssl_profile_count(bigip) == 4)
+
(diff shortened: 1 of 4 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/test/test_ssl_profile.py
```diff
@@ -0,0 +1,119 @@
...
+            mock.MagicMock(return_value=[1, 2, 3, 4])
+        assert(SSLProfileHelper.get_client_ssl_profile_count(bigip) == 4)
+
+    def test_create_client_ssl_profile_exists(self):
+        bigip = mock.MagicMock()
+        bigip.tm.ltm.profile.client_ssls.client_ssl =\
+            mock.MagicMock(return_value=True)
+        rv = SSLProfileHelper.create_client_ssl_profile(
+            bigip, 'testprofile', 'testcert', 'testkey')
+        assert(rv is None)
```
pjbreaux: This is a fine test. We often try to add a little more substance by making assertions about some (or all) the calls made to the mocked object like that in https://github.com/F5Networks/f5-openstack-agent/blob/master/f5_openstack_agent/lbaasv2/drivers/bigip/ssl_profile.py#L38. Just a tip. No need to change anything.

f5_openstack_agent/lbaasv2/drivers/bigip/test/test_ssl_profile.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def test_create_client_ssl_profile_parent_none(self):
+        bigip = mock.MagicMock()
+        bigip.tm.ltm.profile.client_ssls.client_ssl = mock.MagicMock()
+        bigip.tm.ltm.profile.client_ssls.client_ssl.exists.return_value = False
+        SSLProfileHelper.create_client_ssl_profile(
+            bigip, 'testprofile', 'testcert', 'testkey', parent_profile=None)
+        bigip.tm.ltm.profile.client_ssls.client_ssl.create.assert_called_with(
```
pjbreaux: Just like in this test. This makes sure we are calling the mocked bigip object in the way we expected to. We often find surprising things when dealing with complex objects.

PR #327, opened 2016-09-23
#325 tagMode should be specified when creating a vlan interface

### LINKED ISSUE #325: tagMode should be specified when creating a vlan inteface
* Title: VLAN tagMode needs to be specified when creating a VLAN interface if that interface is to be tagged
* Details: The issue is closely related to the an issue defined in f5-sdk, where the tagMode is being required as a creation parameter when creating a vlan interface. For the traceback seen by that user, go here:
https://github.com/F5Networks/f5-common-python/issues/713

#### OpenStack Release
Fix should go into Liberty

#### Agent Version
All versions should be affected

richbrowne:
@jlongstaf 

Fixes #325 

#### What's this change do?
Changes the way in which tagged vlan creation is performed.  Specifying tagMode is required.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_vlan.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_vlan.py
@@ -23,11 +23,6 @@ import mock
 import pytest
 
 
-tagged_vlan_no_int = {'name': "test_valan",
-                      'partition': "Project_123456789",
-                      'tag': 1000}
-
-
 class TestVLANCreate(object):
 
     @pytest.fixture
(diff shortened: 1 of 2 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/test/test_vlan.py
```diff
@@ -0,0 +1,143 @@
...
+        bigip.tm.net.vlans.vlan.exists.assert_called_once_with(
+            name='test_vlan', partition='Project_123456789')
+        bigip.tm.net.vlans.vlan.load.assert_called_once_with(
+            name='test_vlan', partition='Project_123456789')
+
+    def test_create_vlan_no_int(self, bigip, network_helper):
+
+        tagged_vlan_no_int = {'name': "test_vlan",
+                              'partition': "Project_123456789",
+                              'tag': 1000}
```
pjbreaux: Is this structure supposed to have no tag? What does no int mean?

pjbreaux: Ah no interface. I think this test could be clearer with some additional assertions regarding how no interface is created.

f5_openstack_agent/lbaasv2/drivers/bigip/test/test_vlan.py
```diff
@@ -0,0 +1,143 @@
...
+        bigip.tm.net.vlans.vlan.exists.return_value = False
+        bigip.tm.net.vlans.vlan.create.return_value = (
+            bigip.tm.net.vlans.vlan)
+
+        v = network_helper.create_vlan(bigip, vlan_with_untagged_int)
+
+        bigip.tm.net.vlans.vlan.create.assert_called_once_with(
+            name='test_vlan', partition='Project_123456789', tag=0)
+
+        v.interfaces_s.interfaces.create.assert_called_with(
```
pjbreaux: In the previous test, you can assert this function was not called on the v object.

richbrowne:
Changed the test cases to incorporate Paul's feedback.

PR #343, opened 2016-10-03
Start adding modifications to integrate with bb

### LINKED ISSUE #331: This repository must be modified to facilitate buildbot integration

zancas:
@mattgreene 
Fixes #331
...

#### What's this change do?
In order to integrate with the f5pdbld lab build system this repository needs to be modified with the targets expected by buildbot.   The entrypoint for buildbot is "systest/Makefile"

Begin by examining the build targets in `systest/Makefile` they'll indicate which other procedures will be run, and in what order.

#### Any background context?

Before this PR can close, references will have to be updated from POC uri's to production URIs.

Two points that must be addressed in review:

* Is it acceptable to pip install test code?  The modifications to setup.py will cause this to happen.
* Is it acceptable to unpin distribution versions in `requirements.test.txt` again this PR will cause this behavior unless amended.


zancas:
Hmmm...    travis appears to target `F5Networks` where (I think) it should target `zancas`...

zancas:
OK, folks I've repinned requirements.test.txt in accordance with best practice.

richbrowne:
The build failed because of authentication failure to bldr-git.  Is this something that we need to use instead of github?

zancas:
I'm aware of this, I intend to create a `requirements.buildbot.txt` that .travis ignores.  Does that seem like an acceptable solution?

requirements.test.txt
```diff
@@ -1,21 +1,16 @@
...
+f5-sdk==1.5.0
+mock==2.0.0
+pytest==2.9.2
 pytest-cov==2.3.1
-responses
-coverage==4.2
-python-coveralls==2.7.0
+responses==0.5.1
+coverage==4.1
+python-coveralls==1.1
```
mattgreene: Why did you downgrade the the last two requirements?  This seems unrelated to a buildbot integration and may break how we collect code coverage and upload to coveralls.

test/functional/neutronless/l2adjacent/test_disconnected_service_creation.py
```diff
@@ -166,10 +175,13 @@ def setup_l2adjacent_test(request, bigip, makelogdir):
 
 def handle_init_registry(bigip, icd_configuration,
                          create_mock_rpc=create_default_mock_rpc_plugin):
+    print(type(bigip))
+    from pprint import pprint as pp
     init_registry = register_device(bigip)
+    pp(bigip.raw)
```
mattgreene: Are lines 178-179. 191 for debug and can now be removed?  I don't see the value of dumping the raw object, except maybe in case of an error.

zancas:
Roger, addressed.

PR #349, opened 2016-10-05
added static agent ID configuration

### LINKED ISSUE #348: Statically define an agent's Neutron agent ID in the agent config
#### OpenStack Release
Liberty, Mitaka

#### Description
Allow the operator to statically set what agent ID is used in communication and scheduling from Neutron. This would allow agent process mobility between hosts without requiring DB binding remapping. It would also all for dynamic management of agent processes independent from the current hostname and iControl attributes currently used to create the agent ID.

#### Agent Version
8.0.8+, 9.0.3+

#### Operating System
Any

#### OpenStack Release
Liberty, Mitaka+

jgruber:
@mattgreene 
Fixes #348
WIP #348
...

#### What's this change do?
Allows the operator to set a static agent_id setting in the agent ini. If set this setting will override the current agent host name + UUID5 hash of the environment_prefix and first iControl endpoint host.

If not set, the behavior is the same as before this feature.

I also removed some unused agent_host logic in the icontrol driver. 

Build and install LBaaSv2 agent. 

1) Leave agent_id commented out from the agent config and confirm the old agent naming scheme is still working with Neutron. Confirm with neutron agent list.
2) Change the agent_id to a unique fqdn formatted string and restart the agent. Neutron should now see the new agent_id for host scheduling. Confirm with neutron agent list.

#### Any background context?
Customers need this to be able to launch agent processes (in containers or directly on hosts) with process managers, like pacemaker. This feature gives them control over the agent registration.

### DIFF
--- a/etc/neutron/services/f5/f5-openstack-agent.ini
+++ b/etc/neutron/services/f5/f5-openstack-agent.ini
@@ -46,6 +46,13 @@ periodic_interval = 10
 #
 # service_resync_interval = 500
 #
+###############################################################################
+#  Environment Settings
+###############################################################################
+#
+# Since many TMOS® object names must start with an alpha character
+# the environment_prefix is used to prefix all service objects.
+#
 # Objects created on the BIG-IP® by this agent will have their names prefixed
 # by an environment string. This allows you set this string.  The default is
 # 'Project'.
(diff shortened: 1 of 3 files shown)


richbrowne:
John, I just realized that this was merged into master.  Was it your intention to have this available in Liberty/Mitaka?

PR #373, opened 2016-10-07
Update of TLS listener fails

### LINKED ISSUE #372: Update of TLS listener fails
#### OpenStack Release
liberty

#### Description
Updating a TLS listener fails with duplicate default SNI error.

#### Agent Version
8.x

jlongstaf:
@richbrowne 
Fixes #372

#### What's this change do?
Deletes and re-creates virtual server associated with listener when updating a listener.

lbaas_builder.py

#### Any background context?

Issues:
Fixes #372

Problem: Updating a TLS listener fails because of BIG-IP error.

Analysis: Update normally relies on simply updating (PATCH) a
virtual server with changed attributes. For TLS though, the SSL
profile attributes need to be modified as well. To fix, modified
udpate to delete existing virtual server and create a new.
In this way, the SSL profiles are removed and added correctly.

Tests: tempest test suite.


richbrowne:
Reviewed with Jeff.

PR #404, opened 2016-11-03
Implement lbaas-loadbalancer-stats command

jlongstaf:
@richbrowne 
Adding support for lbaas stats

#### What's this change do?
Adds getting VS stats

icontrol_driver.py

#### Any background context?
No

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
@@ -658,6 +658,17 @@ class LbaasAgentManager(periodic_task.PeriodicTasks):  # b --> B
             LOG.error("Exception: %s" % exc.message)
 
     @log_helpers.log_method_call
+    def update_loadbalancer_stats(self, context, loadbalancer, service):
+        """Handle RPC cast from plugin to get stats."""
+        try:
+            self.lbdriver.get_stats(service)
+            self.cache.put(service, self.agent_host)
+        except q_exception.NeutronException as exc:
+            LOG.error("q_exception.NeutronException: %s" % exc.msg)
+        except Exception as exc:
+            LOG.error("Exception: %s" % exc.message)
+
+    @log_helpers.log_method_call
     def create_listener(self, context, listener, service):
         """Handle RPC cast from plugin to create_listener."""
         try:
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
@@ -440,6 +440,25 @@ class LbaasAgentManager(periodic_task.PeriodicTasks):  # b --> B
             # Resync the next time around.
             self.needs_resync = True
 
+    @periodic_task.periodic_task(spacing=30)
+    def collect_stats(self, context):
+        if not self.plugin_rpc:
+            return
+
+        active_loadbalancers = (self.plugin_rpc.get_active_loadbalancers())
+        for loadbalancer in active_loadbalancers:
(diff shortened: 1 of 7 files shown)


jlongstaf:
@swormke: I will create a new test to validate the stats.

jlongstaf:
@swormke: As requested, added Tempest scenario test to replace test_solution.py. See test_stats.py in driver PR: https://github.com/F5Networks/f5-openstack-lbaasv2-driver/pull/340

richbrowne:
We need to figure out what to do about the change in the stats API between BIG-IP 11.6.x and 12.1.  I would suggest doing some runtime check to determine the what version we are running.  Unfortunately, this is not very clean.

jlongstaf:
I’d rather avoid having to do version checks. I have a fix that looks like this. I haven’t tested this at all, so I don’t know if it is correct.

    def get_stats(self, bigip,  name=None, partition=None, stats=[]):
        collected_stats = {}

        # get resource, then its stats
        if self.exists(bigip, name=name, partition=partition):
            resource = self.load(bigip, name=name, partition=partition)
            resource_stats = resource.stats.load()
            stat_entries = resource_stats.entries

            # Difference between 11.6 and 12.1. Stats in 12.1 are embedded
            # in nestedStats. In 11.6, they are directly accessible in entries.
            if stats[0] not in stat_entries:
                # find nestedStats
                for key in stat_entries.keys():
                    value = stat_entries.get(key, None)
                    if 'nestedStats' in value:
                        stat_entries = value['nestedStats']['entries']


            # add stats defined in input stats array
            for stat in stats:
                if stat in stat_entries:
                    collected_stats[stat] = \
                        stat_entries[stat]['value']

richbrowne:
Hi Jeff,

Ok, that sounds like good approach.

jlongstaf:
Tested last set of changes with both 11.6 and 12.1 versions of BIG-IP. Ready for final review.

f5_openstack_agent/lbaasv2/drivers/bigip/resource_helper.py
```diff
@@ -244,3 +244,43 @@ def _collection(self, bigip):
                       "resource %s", self.resource_type)
             raise KeyError("No collection available for %s" %
                            (self.resource_type))
+
+    def get_stats(self, bigip,  name=None, partition=None, stats=[]):
```
richbrowne: Call stats, statKeys, this way we know we are searching for something and not passing in statistics.  Nit-picky, but it might help readability.

f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
```diff
@@ -440,6 +440,25 @@ def periodic_resync(self, context):
             # Resync the next time around.
             self.needs_resync = True
 
+    @periodic_task.periodic_task(spacing=30)
```
richbrowne: The spacing here seems a little too frequent; however, instead of pushing stats to neutron, maybe we should just pull from the Neutron API.  Only populate stats when a call to get_loadbalancer_statistics is made.

f5_openstack_agent/lbaasv2/drivers/bigip/resource_helper.py
```diff
@@ -244,3 +244,43 @@ def _collection(self, bigip):
...
+        """
+        collected_stats = {}
+
+        # get resource, then its stats
+        if self.exists(bigip, name=name, partition=partition):
+            resource = self.load(bigip, name=name, partition=partition)
+            resource_stats = resource.stats.load()
+            stat_entries = resource_stats.entries
+
+            # Difference between 11.6 and 12.1. Stats in 12.1 are embedded
```
jlongstaf: OK. I'll open a new issue for this and leave the code as it is.

PR #409, opened 2016-11-08
Add service adapter for l7policies and l7rules.

### LINKED ISSUE #408: Add service adapter for l7policies and l7rules.
#### OpenStack Release
Starting with mitaka

#### Description
In order to support L7 Content Switching in the agent, we must be able to adapt the policies and rules in the service object to conform to the bigip's view of these objects. This is the realm of the service adapter. This adapter is a class, instead of a class method as before, that inherits from ServiceModelAdapter and manages the translation.

pjbreaux:
@jlongstaf 

Fixes #408 

#### What's this change do?
Add service adapter for l7policies and l7rules.

Start in the new l7policy_adapter module.

#### Any background context?
In order to support L7 Content Switching in the agent, we must be able to adapt the policies and rules in the service object to conform to the bigip's view of these objects. This is the realm of the service adapter.

There is also a doc in the new tests to show you how to run them. Nothing really new, just creating a symbols file and doing py.test ...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_adapter.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_adapter.py
@@ -1,0 +1,156 @@
+# coding=utf-8
+# Copyright 2014-2016 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_adapter.py
```diff
@@ -0,0 +1,156 @@
...
+        if not self.service['l7rules']:
+            msg = 'No Rules given to implement. A Policy cannot be attached ' \
+                'to a Virtual until it has one or more Rules.'
+            raise PolicyHasNoRules(msg)
+        self.policy_dict = {}
+        self.policy_dict['name'] = 'wrapper_policy'
+        self.policy_dict['partition'] = self.folder
+        self.policy_dict['strategy'] = 'first-match'
+        self.policy_dict['rules'] = []
+        self.policy_dict['legacy'] = True
```
jlongstaf: The legacy attribute causes a failure w/ 11.6. -- unexpected json pair.

PR #416, opened 2016-11-17
Feature.rename service objects

richbrowne:
@pjbreaux @jlongstaf @zancas 

   The migration from using OpenStack names to name LTM objects
    on the BigIP to the OpenStack UUIDs requires a test of the
    translation from old to new names of existing LBaaS services.
    This commit creates a docker container and test that will run
    a neutronless agent that runs the legacy implementation.  Upgrade
    tests can be executed using this container to deploy services
    with the old algorithm.  The tests code for upgrade is in
    this commit, but the full run_tests entry point needs to be
    updated to execute the rename and teardown.


richbrowne:
Ok'd by Jeff L

PR #436, opened 2016-11-23
pin neutron to that branch implied by subject code

### LINKED ISSUE #450: The version of neutron-lbaas used by nightly regression runs is incorrectly set.
* Title: This issue is a bug, neutron-lbaas should be pip installed as "@stable/$branch" to match the version of `f5-openstack-agent` we are running against.

zancas:
@mattgreene 
Pinning the correct version of neutron is critical to regression stability...  this is particularly the case because back-compatibility is not guaranteed. 

Fixes #450

### DIFF
--- a/systest/scripts/install_systests.sh
+++ b/systest/scripts/install_systests.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -ex
 
+if [ "$1" == "-h" ] || [[ "$1" == "--help" ]]; then
+    echo "Usage: This script expects to be called from the buildbot interface (systest/Makefile or systest/????.yml).  Appropriate arguments are established in that context."
+fi
 if [ "$1" == "" ]; then
     echo "ERROR: no repo value provided!"
     exit 1
@@ -34,7 +37,7 @@ cd f5-openstack-agent
 pip install --upgrade .
 
 # Install openstack deps (if necessary)
-pip install git+https://github.com/openstack/neutron@9.1.0
+pip install git+https://github.com/openstack/neutron@stable/$branch
 pip install git+https://github.com/openstack/oslo.log.git@stable/$branch
 pip install git+https://github.com/openstack/neutron-lbaas.git@stable/$branch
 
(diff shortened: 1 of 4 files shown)


zancas:
Obviously we should validate that the regression (neutronless etc.) tests run before accepting this PR.

That _should_ be about as hard as running `cd systest && make` provided the test system has appropriate routes and ssh keys.    I say _should_ because I haven't been actively testing these suites for several weeks, and code has been changed since then.

I'll try to check in tomorrow.

swormke:
A couple of comments.

1. Create a proper bash usage statement for this script that is output after the error messages so people know how to use it, and you probably want to accept a -h or --help option as well that just prints that usage statement and exits.

1. Remember that all commits must have an issue, make sure you have a issue for this problem in the repo and a proper commit message before this is merged in. Please refer to https://github.com/F5Networks/f5-openstack-agent/blob/master/.git-commit-template.txt if you need a refresher on what that looks like.  I generally add this to my .gitconfig so that each time I commit I have the right template.  Something like this.
`
$> git config commit.template relative-path-from-repository-root
`

mattgreene:
This PR needs to be cleaned up => git rebase to pull in latest changed to liberty.  The recent commit from @richbrowne should not show up in this PR.

zancas:
I investigated this CI failure here's the error:

'Error: Cannot retrieve metalink for repository: epel. Please verify its path and try again'


It's in the RPM packaging section.   This seems unlikely to have been caused by my changes to test code.

zancas:
The parent section of the CI fail, error message:

'The command "f5-openstack-agent-dist/scripts/package_agent.sh "redhat" "6"" exited with 1.'

zancas:
As noted above there are sporadic REL 6 packing errors in the CI system. "Fixed" by rerunning the build.

zancas:
I intend to push my fix for #463 on top of this commit, so it'd be a convenience if it merge first.

mattgreene:
A couple of additional comments:

1. I don't see that this PR references a github issue- please create one.

2. Pinning the branch correctly is required.  If that introduces some intermittent RPM issue, that also needs to be fixed but can go under another PR.  It is not a solution to claim the issue is "fixed" by rerunning the build.

3. I don't follow your last comment about getting this PR landed before fixing #463.  Does this fix change or eliminate that other bug?  Generally, there should be 1 change-set:1 bug:1 PR.

dflanigan:
All Matt's comments have been addressed.

PR #455, opened 2016-12-01
Disconnected network tunnel remains after deleting a loadbalancer

### LINKED ISSUE #293: Delete Loadbalancer when run against disconnected service-supporting agents fails
@mattgreene 
Possibly I should be responsible for fixing this, but I'd like you to sanity check me on this one!
- Title: When I run `nclientmanager.delete_all_loadbalancers` against `disconnected_service` supporting agents the test partition is not removed.
- Details: 
  In the `test_disconnected_service_creation.py` agentless tests I implemented cleanup of FDB tunnels here:

https://github.com/F5Networks/f5-openstack-agent/blob/55822c421c02299bd909fa84afdd4d02f13df900/test/functional/neutronless/conftest.py#L71

this caused the tests to reset the device to a clean state without validating that the agent "delete" function.   This is acceptable since the tests were not of the `delete` functionality, but it would be better to make the tests use/depend on the delete functionality implemented within the agent itself.
#### Agent Version

8.0.7
#### OpenStack Release

Liberty

richbrowne:
@mattgreene 
Fixes #293

#### What's this change do?
Fixes the bug where the disconnected network hung around
after a loadbalancer was deleted.

The disconnected_network and tenant modules handle the
disconnected_network lifecycle.

#### Any background context?
This bug is easily reproducible in GRM, but it it not clear if this
bug was encountered in under cloud deployments.  The tests
deploy a loadbalancer and delete it in global routed mode.

Issues:
Fixes #293

Problem:
When creating and deleting a loadbalancer in global routed mode,
the tenant folder is not deleted because the disconnected network
is hanging around.

Analysis:
When the folder cleanup is invoked the delete_network() for
disconnected network is only called when there are route domains
present in the folder; otherwise, the deletion of disconnected
network is skipped.

Code inspection revealed that the create_network and delete_network calls
in disconnected_service are not symmetric.

create_network() invokes network_helper.create_tunnel whereas
delete_network() directly loads the tunnel object via bigip.tm
and does a .delete().

Tests:
    test_disconnected_service_creation.py::test_featureoff_withsegid_lb PASSED
    test_disconnected_service_creation.py::test_withsegid_lb PASSED
    test_disconnected_service_creation.py::test_featureoff_withsegid_listener PASSED
    test_disconnected_service_creation.py::test_featureoff_nosegid_lb PASSED
    test_disconnected_service_creation.py::test_featureoff_nosegid_listener PASSED
    test_disconnected_service_creation.py::test_withsegid_listener PASSED
    test_disconnected_service_creation.py::test_nosegid_lb PASSED
    test_disconnected_service_creation.py::test_nosegid_listener PASSED
    test_disconnected_service_creation.py::test_nosegid_listener_timeout PASSED
    test_disconnected_service_creation.py::test_nosegid_to_segid PASSED
    test_disconnected_service_creation.py::test_featureoff_grm_lb PASSED


mattgreene:
@richbrowne Thanks for the detailed PR write-up and new test to repro and prove out the fix.

PR #468, opened 2016-12-06
The devtools directory in the metasource directory is obsolete, remove devtools

### LINKED ISSUE #461: Remove unused and stale devtools directory
The [devtools](devtools) directory was created prior to standardizing our test environment. We have now standardized on [tox](http://tox.readthedocs.io/en/latest/) as our test runner and we will no longer work to keep the docker files and execution methods recorded in devtools up to date.

#### OpenStack Release
Liberty

#### Description
The devtools directory contains environment management and docker files that are not directly related to the success and testing of the project. Docker can still be used as a a tool for running the tests, but we must decouple the infrastructure and environment used to execute the tests from the tests and repo itself.

The expectation going forward is that users will use tox to execute whether on a workstation, VM, container etc...

zancas:
@swormke @richbrowne @pjbreaux @jlongstaf @mattgreene @szakeri

What issues does this address?

Fixes #461

I've validated that removing this directory does not break the regression tests.

I've included all stake holders in case someone has critical code in that directory. If you do, I doubt we should not merge this, rather you should extract that code (e.g. with cherry-pick) from "the past" and add it back to the repo as appropriate. Nevertheless, I didn't want anyone to be shocked to discover that directory is gone.

It'd be nice if we could get acks from stakeholders before merging.


mattgreene:
Shawn's PR will resolve this.  It will merge shortly after fixing a merge conflict.

PR #471, opened 2016-12-08
Makefile targets need to use virtualenv for internal testing #470

### LINKED ISSUE #470: Makefile targets need to use virtualenv for internal testing
#### Description

A test that is run via the Makefile for internal testing should use a virtual environment so that it can install some of the internal test packages and not conflict or update the base system.  It should also run the tools inside that virtual environment.

#### OpenStack Release
Liberty+

swormke:
@mattgreene 
### Issues
Fixes #470

### Problem
A test that is run via the Makefile for internal testing should use a virtual environment so that it can install some of the internal test packages and not conflict or update the base system. It should also run the tools inside that virtual environment.

### Analysis
* Update the make file to create a virtualenv
* Use the virtualenv for the commands that need it
* Add a clean target to remove test artificats in tox and make

### Tests
* make functest

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -30,7 +30,7 @@ VENV_ACTIVATE := $(MAKEFILE_DIR)/$(VENV)/bin/activate
 
 # Before we run any tests we need a virtualenv and a few packages
 pre-build:
-	virtualenv buildbot; \
+	virtualenv $(VENV); \
 	. $(VENV_ACTIVATE); \
 	pip install tox; \
 	pip install git+ssh://git@bldr-git.int.lineratesystems.com/tools/testenv.git; \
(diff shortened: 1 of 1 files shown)


systest/Makefile
```diff
@@ -24,14 +25,27 @@ disconnected_results := $(RESULTSDIR)/$(disconnected_session)
...
 testenv_config := bigip.testenv.yaml
 
-unit:
-	cd $(MAKEFILE_DIR)/../
+VENV := buildbot
+VENV_ACTIVATE := $(MAKEFILE_DIR)/$(VENV)/bin/activate
+
+# Before we run any tests we need a virtualenv and a few packages
+pre-build:
+	virtualenv buildbot; \
```
mattgreene: Replace 'buildbot' with $(VENV)

Issue #381, opened 2016-10-10
Test vcmp has hardcoded 12.0 version values in the tests and test files


pjbreaux:
- Title: 
- Details: The VCMP neutronless tests are inflexible to testing multiple versions of TMOS. We need to have the test runner provide a release (tmos version) when the test is run, so the test can know what to expect for tmos version. This exists in two places: one is in the vcmp_uris.json file and the other is in the test, where the vcmp guests has names like bigip-12.0. Those need to be more generic.

dflanigan:
The merge comment did not reference this issue correctly and could not close this issue automatically. (And I observed the commits in master.)

PR #472, opened 2016-12-08
Deleting last VLAN in VCMP scenario results in all VLANs being removed from guest

### LINKED ISSUE #447: Deleting last VLAN in VCMP scenario results in all VLANs being removed from guest
In a vCMP enabled scenario we are seeing an issue that when the last load balancer in the system is deleted then all VLANs assigned to the VCMP guest are removed, including those not managed but the Openstack agent. 

This leads to a situation where the management VLAN is removed and the guests are left in a disconnected state.

pjbreaux:
@jlongstaf @dflanigan 

Fixes #447 and #381 

#### What's this change do?
Found that the VLANs were being wiped by a bad modify call in the vcmp
module. Changed the code to capture the existing vlan list, then remove
the element we want, and perform the modify call.

Code change is rather minor. The other changes in the vcmp module should only be related to log output. This is for more consistent output messages.

The tests changed quite a bit, in that they no longer have hard-coded values for our vcmp hosts and guests. The parameters are now all fed in from the test environment file.

#### Any background context?
In a vCMP enabled scenario we are seeing an issue that when the last
load balancer in the system is deleted then all VLANs assigned to the
VCMP guest are removed, including those not managed but the Openstack
agent.

This leads to a situation where the management VLAN is removed and the
guests are left in a disconnected state.


pjbreaux:
@dflanigan: I had check look at the code change, and I'd like you and I to walk through the test change, and have you actually run the tests, if you have time.

pjbreaux:
Dennis and I ran these tests multiple times and the first time Dennis ran, one of the tests hung. We never saw it give up, so we don't know the actual failure. After rerunning multiple times, we saw nothing but passes for all tests.

PR #473, opened 2016-12-08
Unable to delete pools with shared nodes (IP addresses)

### LINKED ISSUE #469: Unable to delete pools with shared nodes (IP addresses)
#### OpenStack Release
Mitaka

#### Agent Version
9.1.0

#### Operating System
Suse

#### OpenStack Release
Mitaka

#### Description
Ceating a loadbalancer with two listenrs, two pools (not shared) and two members, but with the members having the same IP address. Unable to delete pool. Neutron reports that it deleted pool but the pool remains on BIG-IP and in Neutron database.

#### Deployment
See above.

jlongstaf:
@richbrowne 
Fixes #469 

#### What's this change do?
Adds exception handling when deleting a member node. 

pool_service.py

#### Any background context?
Original code does handle use case where a node is shared between members.

Issues:
Fixes #469

Problem: BIG-IP returns an error when deleting a node
that is shared between two or more members.

Analysis: Added exception handling to look for 400 status code
and ignore, assuming that node is shared.

Tests:

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/pool_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/pool_service.py
@@ -15,6 +15,7 @@
 #
 
 from oslo_log import log as logging
+from requests import HTTPError
 import urllib
 
 from f5_openstack_agent.lbaasv2.drivers.bigip.resource_helper import \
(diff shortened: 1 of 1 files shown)


richbrowne:
Performed manual test of this fix.  I believe it should be moved forward while I automate the test.

PR #527, opened 2017-01-13
the f5_os_test package brought overhead no benefit

### LINKED ISSUE #526: The f5_os_test repository costs dependency overhead for little benefit
We need to move its logic into the relevant source repos (e.g. `f5-common-python`) and remove it.
This will simplify dependency management which is significantly beneficial.

zancas:
Issues:
Fixes #526, #548

Problem:  Logic was innappropriately bundled into a separate
repository "f5-openstack-test" this caused maintenance,
packaging, and dependency management issues for little benefit.

Analysis:  The logic required by f5-openstack-agent has now
been migrated into f5-common-python (upon which it already depends)

NOTE this PR cannot be closed until the 2.2.0 version of the SDK is
released.

@richbrowne 
Fixes #526, #548

...

#### What's this change do?
Completely eliminate reference to f5-openstack-test.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,2 @@
 [bdist_rpm]
-requires =  f5-sdk == 1.5.0
+requires =  f5-sdk >= 2.2.0
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ __pycache__/
 *.so
 
 # Distribution / packaging
+.containenv
 .Python
 env/
 build/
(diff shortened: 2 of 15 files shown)


setup.cfg
```diff
@@ -1,2 +1,2 @@
 [bdist_rpm]
-requires =  f5-sdk == 1.5.0
```
richbrowne: Since we are only using version 2.2 for the test, do we really need to update the version?

PR #567, opened 2017-01-30
ESD tag processor implementation 

### LINKED ISSUE #566: ESD tag processor
#### OpenStack Release
Mitaka

#### Description
Enhancement required for ESD tag processor.  This enhancement will create a mapping between a valid tag and the bigip resource type and will check if the esd tag that user created exists on bigip.

szakeri:
@jlongstaf 
Fixes #566 

#### What's this change do?
This change will create a valid set of tags by returning intersection of known valid ESD tags and the user tags. Then based on the mapping of ESD tags to bigip resources, it will validate if the resource exists on bigip. Also it will check if a tag is valid or has a valid value type.

EsdTagProcess class inside esd_filehandler module


jlongstaf:
Code is fine  and we will fix Travis build failures later.

PR #578, opened 2017-02-08
Issues: 418 - addresses issue of init.d not behaving the way as expected

### LINKED ISSUE #418: rpm packages's init.d script is not compatible with Suse Cloud 6
Hi 

It appears that Suse lacks the the log_daemon_msg and start-stop-daemon commands hence the /etc/init.d/f5-oslbaasv2-agent doesn't work:

Nov 17 08:36:33 dc8-1f-66-b8-bd-ca f5-oslbaasv2-agent[42674]: /etc/init.d/f5-oslbaasv2-agent: line 124: log_daemon_msg: command not found
Nov 17 08:36:33 dc8-1f-66-b8-bd-ca f5-oslbaasv2-agent[42674]: /etc/init.d/f5-oslbaasv2-agent: line 71: start-stop-daemon: command not found
Nov 17 08:36:33 dc8-1f-66-b8-bd-ca f5-oslbaasv2-agent[42674]: /etc/init.d/f5-oslbaasv2-agent: line 136: log_end_msg: command not found

ssorenso:
Fixes #418 

Problem:
The issue of the init.d script not exiting as expected with the correct,
printed output into the CLI for start, stop, restart.  Have already
fixed the status in the past.

Analysis:
Am introducing additional logic to test the status of the daemon prior
to any start or stop.  If the daemon was already running, then it will
simply state that it was&is running without 'trouncing' the original
PID.
If it was already stopped when 'stop' is executed, then it will state
that it was already in a stopped state and state that it is stopped.  If
it was running when stop is issued, then it will simply state that it is
stopped or the particular error coded type, and state that it is still
running (if so).

Lastly, if the daemon is called from an unrecognized argument,
unprivilidged
user or the improper number of arguments are given, then it will exit
with the
proper exit statement and code.

Tests:
Have tested the following manually as this requires a full devstack or
similar build on ubuntu:
- start; restart; stop
$ sudo service f5-oslbaasv2-agent start;sudo service f5-oslbaasv2-agent
restart;sudo service f5-oslbaasv2-agent stop
(0) Service is running
(0) Service is running
(7) Service, f5-oslbaasv2-agent, is not running
- from stop, status; start; status; #kill# status; stop; start; start
$ sudo service f5-oslbaasv2-agent status
(3) Service f5-oslbaasv2-agent is not running!
$ sudo service f5-oslbaasv2-agent start
(0) Service is running
$ sudo service f5-oslbaasv2-agent status
(0) Service f5-oslbaasv2-agent is in an OK Status!
$ kill -9 $(cat /var/run/neutron/f5-oslbaasv2-agent.pid)
$ sudo service f5-oslbaasv2-agent status
(1) Service f5-oslbaasv2-agent is dead and /var/run pid file exists!
$ sudo service f5-oslbaasv2-agent stop
Service is already in stopped status
(7) Service, f5-oslbaasv2-agent, is not running
$ sudo service f5-oslbaasv2-agent start;sudo service f5-oslbaasv2-agent
start
(0) Service is running
Service is Already Running: no action
(0) Service is running

@richbrowne 
Fixes #418 
WIP #
...

#### What's this change do?
Please see above in commit

The ./etc/init.d/f5-oslbaasv2-agent file

#### Any background context?
Please refer to the Linux's standard handling of init.d files for handling start-stop-daemon.

### DIFF
--- a/etc/init.d/f5-oslbaasv2-agent
+++ b/etc/init.d/f5-oslbaasv2-agent
@@ -2,7 +2,7 @@
 ### BEGIN INIT INFO
 # Provides:             f5-openstack-agent
 # Required-Start:       $remote_fs $syslog
-# Required-Stope:       $remote_fs $syslog
+# Required-Stop:        $remote_fs $syslog
 # Should-Start:         keystone neutron-server
 # Should-Stop:          keystone neutron-server
 # Default-Start:        2 3 4 5
@@ -20,6 +20,7 @@ F5_AGENT_CONF="/etc/neutron/services/f5/f5-openstack-agent.ini"
 STARTDAEMON_ARGS=""
 DEPEND_RETURN=''
 DEPEND_EXIT=''
+ACTION=$1
 
 [ -r "${F5_AGENT_CONF}" ] && DAEMON_ARGS="${DAEMON_ARGS} --config-file ${F5_AGENT_CONF}"
 [ -r "${NEUTRON_CONF}" ] && DAEMON_ARGS="${DAEMON_ARGS} --config-file ${NEUTRON_CONF}"
@@ -47,7 +48,7 @@ if [ -z "${SYSTEM_USER}" ]; then
 fi
 
 PIDFILE="/var/run/${PROJECT_NAME}/${NAME}.pid"
-STARTDAEMON_ARGS=" --start  $STARTDAEMON_CHUID --background --make-pidfile --pidfile ${PIDFILE}"
+STARTDAEMON_ARGS=" --start  $STARTDAEMON_CHUID --make-pidfile --pidfile ${PIDFILE}"
 [ -x $DAEMON ] || exit 4
 
 # If ran as root, create /var/lock/X, /var/run/X, /var/lib/X and /var/log/X as needed
@@ -69,7 +70,7 @@ LOGFILE="/var/log/${PROJECT_NAME}/${NAME}.log"
 [ "x$USE_LOGFILE" != "xno" ] && DAEMON_ARGS="$DAEMON_ARGS --log-file=$LOGFILE"
 
 function log_initializing_msg() {
-    msg="Initializing ${ACTION} on ${NAME}..."
+    msg="Begin ${ACTION} on ${NAME}..."
     logger "${msg}"
     echo "${msg}"
 }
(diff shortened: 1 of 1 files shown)


richbrowne:
Ok, try the following:

Edit the configuration file
/etc/neutron/services/f5/f5-openstack-agent.init by putting a space before
any option.  Try periodic interval.  Run the script and see if you get the
error.

Rich

ssorenso:
Does it make sense to:

1. Store the STDOUT/STDERR into a variable
2. Reference the return code from start-stop-daemon
3. Dump into the terminal upon error or...
4. Remain quiet upon success

Users can then go without the white noise when things are 'okay' and see the dumped terminal output upon discovery of an error.

I am still using the start-stop-daemon executable quite heavily in that it...

- Still initiates and stops the service
  - Handling the PID and LOCK files accordingly
- Use it to check the status to report appropriately if/when circumstances are abnormal
  - When the service is running with what items present

If I were to rewrite it, though I would suggest in Perl.  I would not normally suggest this; however, given how complicated the parsing is to grab things, Perl is a bit more stable with log handling than straight shell.  Even bash, which is quite handy and can handle a good number of things.  Further, it would greatly reduce the number of lines we already have in bash as there are multiple steps to compensate for the lack of intelligence here.

Python is a lot of overhead to start/stop a service; however, it can be used to replace this as well.  I'm just worried at that point that it's the equivalence of a sludge hammer for ants...

ssorenso:
Have performed the following changes:

- Remove the `--background` flag
- Pipe the `start-stop-daemon` output into a truncated, temp var
  - Contains only trace, error, and critical strings
  - If this causes issues down the road, we can always use like a 40K-60K cache
  - Upon error, dump the piped content

This gives the user:

- What happened and the choice to dump the material into a log file of their own
- An umbrella "catch net" for errors that the service might run into if/when an unhandled exception occurs
- If they want to customize whether or not they want the feedback in the terminal, there's a comment on the line that dumps it in.

ssorenso:
So, I've got a fix for this that should address most of our concerns:

1. I create the $LOCKFILE
2. I cast the `start-stop-daemon call 2>&1 >> $LOGFILE;rm -rf $LOCKFILE` into a string
3. I call `nohup $str >/dev/null 2>&1 &`

What this does is:

* Stores the STDERR/STDOUT into the `/var/log/neutron/f5-oslbaasv2-agent` log
* Gives the running prompt back to the user
* Forks the service job off using `nohup` followed by a `disown`
* Destroys the lock file if something "bad" happens to the agent's service PID during runtime

This keeps the user form having the running instance going in a prompt that they have to then disown while still preserving the STDERR

Issue #597, opened 2017-02-27
Nightly builds need to run against 12.1.1


zancas:

dflanigan:
fixed and working

Issue #596, opened 2017-02-27
Nightly builds need to run against 11.5.4


zancas:

dflanigan:
fixed and working

PR #598, opened 2017-02-27
Add device multiplex to nightly tests

### LINKED ISSUE #594: Nightly builds need to run against 11.6.0

zancas:
Issues:
Fixes #594 #596 #597

Problem: We support a range of devices.  Our nightly infrascture
needs to automatically test each of these devices.

Analysis: I added the changes to the Makefile following the
architecture swormke used in the f5-openstack-lbaasv2-driver repo
which was itself following the practices developed by the velcro
team.   The entrypoint into the nightly build process is that
the buildbot worker container runs (the equivalent of):

cd f5-openstack-agent/systest && make functest

Tests:  The new code was tested by running the above in an
appropriate environement (testenv generated stack on the "lab"
openstack cloud).  This correctly caused 4 separate runs of the
same tests.  Undercloud tests were not relevant for these tests
so the 4 undercloud RULEs simply echoed their names.

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -1,119 +1,184 @@
-.PHONY: all clean pre-build unit systest functest disconnected_service \
-disconnected_service-setup \
-disconnected_service-install disconnected_service-run \
-disconnected_service-teardown singlebigip \
-singlebigip-setup \
-singlebigip-install singlebigip-run \
-singlebigip-teardown
+# Copyright 2016-2017 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
 
-PROJECT := f5-openstack-agent
-repo := https://github.com/F5Networks/$(PROJECT).git
-ssh_conf := testenv_symbols/testenv_ssh_config
+#
+# This file is used by the F5 OpenStack team as an interface to the automated
+# test system. It is divided into 4 sections:
+# VARIABLES
+# INITIALIZATION
+# MULTIPLEXING
+# TESTSBYINFRA
 
+#### VARIABLES SECTION:
+#    DEVICES: (See: https://support.f5.com/csp/article/K8986)
+#     A list of MAJOR_MINOR_MAINTENANCE, adding a new triple XX.Y.Z causes
+#     the system to look up the "current actual" image in the
+#     "device_version_maps.sh" file.  For a given XX.Y.Z device the system
+#     tests against a single (OpenStack Ready) image listed in that file.  The
+#     assumption is that the XX.Y.Z triple specifies a hardware platform, and
+#     futher verison info denotes software _for_ that hardware.
+#    DEPLOYS are a list of XX.Y.Z triples combined with other variables,
+#     currently the only specified variable is the (implicit) "CLOUDTYPE" with
+#     the states "overcloud" and "undercloud", therefore elements of DEPLOYS
+#     look like the following: '12_1_1_overcloud' '11_5_4_undercloud'.
+# 
+DEVICES := 11_5_4 11_6_0 11_6_1 12_1_1
+DEPLOYS := $(foreach v, $(DEVICES), $(v)-overcloud $(v)-undercloud)
+
+# Tell make that targets don't map to files
+.PHONY: functest \
+        $(DEPLOYS) \
+        run_overcloud_tests \
+        run_undercloud_tests \
+        singlebigip \
+        setup_singlebigip_tests \
+        run_singlebigip_tests \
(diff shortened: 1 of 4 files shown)


systest/Makefile
```diff
@@ -1,119 +1,178 @@
...
-		--autolog-session $(unit_session) \
-
-functest:
-	$(MAKE) -C . unit
-	$(MAKE) -j -C . functest_all
-
-functest_all:
-	@echo "automated functional tests..."
-	$(MAKE) -C . disconnected_service
+		--autolog-outputdir $${RESULTS_DIR} \
```
mattgreene: RESULTS_DIR is both a makefile and exported env variable?  Suggest picking a different name for one of them so it's obvious when you choose one over the other.

systest/device_version_maps.sh
```diff
@@ -0,0 +1,18 @@
...
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+export BIGIP_11_5_4="bigip-osready-11.5.4.2.0.291"
+export BIGIP_11_6_0="bigip-osready-11.6.0.0.0.401"
+export BIGIP_11_6_1="bigip-osready-11.6.1.1.0.326"
+export BIGIP_12_1_1="bigip-osready-12.1.1.2.0.204"
```
mattgreene: Looks like only the 11.6.1 image exists publicly in the lab openstack deployment.  All images must be present before merging this PR.

swormke:
Do the targets in `#### TESTSBYINFRA SECTION` still work as they intended? Or are they run in the nightly?  The reason I am asking is because of the `$TEST_RESULTS` comment that Matt has above.  It looks like those are just dumping their results into the `./test_results/` directory where as the $(DEPLOYS) targets are dumping them into a deeper directory structure.

```
export GUMBALLS_PROJECT=$(PROJNAME)_$(BRANCH)-$@ &&
export RESULTS_DIR=$(RESULTS_DIR)/$${GUMBALLS_PROJECT} &&
```

Just want to make sure that the results/projects structure is right for these tests after the changes for the multiple deployment options.

swormke:
@mattgreene does this really need over/undercloud iterations? Seems like the issues listed above only talk about versions and the internal Agile board seems to do the same in the headline but the DoD indicates something different. It is fine either way, just wanted to make sure that we are not chewing up cycles at night for topologies that are not required.

systest/Makefile
```diff
@@ -1,119 +1,178 @@
...
+#  This rule sets variables to values specific for a particular deploy. This
+#  is where interfacing software (trtl) names are set.
+#  "GUMBALLS_PROJECT" maps to a raw in gumballs (URL)XXX
+#  "GUMBALLS_SESSION" maps to a column in gumballs" (URL)XXX
+$(DEPLOYS):
+	-@echo executing $@
+	. $(MAKEFILE_DIR)/device_version_maps.sh &&
+	export DEVICE=BIGIP_`python -c 'print("$@".split("-")[0])'` &&
+	export CLOUD=`python -c 'print("$@".split("-")[1])'` &&
+	export GUMBALLS_PROJECT=$(PROJNAME)_$(BRANCH)-$@ &&
```
pjbreaux: Looks like the gumballs project is being named the same as the automatic variable for the target name. That target name is using hyphens instead of underscores. Since these names are important, any reason why we are deviating from what driver is doing?

systest/Makefile
```diff
@@ -1,119 +1,178 @@
...
-disconnected_service-teardown singlebigip \
-singlebigip-setup \
-singlebigip-install singlebigip-run \
-singlebigip-teardown
-
-PROJECT := f5-openstack-agent
-repo := https://github.com/F5Networks/$(PROJECT).git
-ssh_conf := testenv_symbols/testenv_ssh_config
-
+# Copyright 2016 F5 Networks Inc.
```
zancas: Fixed.

systest/Makefile
zancas: Fixed, made variable "AUTOLOGDIR"

systest/Makefile
swormke: @mattgreene sounds reasonable to me.

systest/device_version_maps.sh
zancas: I think you were looking at the deprecated kilo cloud.

zancas:
Per @swormke 's comment, the difference between the output in the "functest" RULE and the "$(DEPLOYS)" RULEs is that the "functest" RULE only runs unit tests.   These tests should not vary when the device or cloud changes, so those tests do not include that information in their GUMBALLS PROJECT.

If a test is labelled "unit" and it is sensitive to one of those variables it should be renamed (moved) into a functional test.

zancas:
@pjbreaux example GUMBALLS_PROJECT names:

`disconnected_singlebigip_f5-openstack-agent_liberty-12_1_1-overcloud`

`singlebigip_f5-openstack-agent_mitaka-11_6_0-overcloud`

`f5-openstack-agent_mitaka-unit`

NOTE:  "unit" GUMBALLS_PROJECTs do not include device version or cloud type as this information shouldn't be relevant to unittests

example GUMBALLS_SESSION (parsed apart by gumballs) names:

`v8.0.1-434-g5d0c106_20170301-204408`

mattgreene:
I'm accepting the PR per verbal approval from @richbrowne.

PR #606, opened 2017-03-02
Opflex member can get stuck in PENDING if operations time-out

### LINKED ISSUE #589: Opflex member stuck in 'PENDING_CREATE' if port creation times out.
If a port member creation fails after a timeout the loadbalancer gets put in ‘ERROR’ state, but the member is left in ‘PENDING_CREATE’.  The loadbalancer should actually be left alone and the member set to ‘ERROR’ provisioning status.  I can fix this, but we wanted to get this to Cisco to move forward.

richbrowne:
Issues:
Fixes #589

Problem:
If a port member creation fails after a timeout the loadbalancer
gets put in ‘ERROR’ state, but the member is left in ‘PENDING_CREATE’.
The loadbalancer should actually be left alone and the member set to
‘ERROR’ provisioning status.

Analysis:
Fix the agent_manager handling of pending services to update the
status of the service in the Neutron.  Properly handle exceptions
of objects that are deleted when exceptions occur in the network
setup.

Tests:
test/functional/neutronless/loadbalancer/
test/functional/neutronless/disconnected/
test/functional/neutronless/member/

Manual testing consisted of:
Create loadbalancer service with all objects: listener, pool, members
Create member with and without existing port.
Create member on opflex, add vlan child segment manually and verify
that the member is connected.
Create loadbalancer on opflex network, add vlan child segment
manuall and verify that the virtual get's created.
Create loadbalancer on opflex network, don't add segment and check
that the loadbalacner goes to 'ERROR' state and that it can be
deleted.
Create a member on the opflex network, don't add segment and check
that the member goes to 'ERROR' state after timeout period and verify
it can be deleted.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
@@ -1409,9 +1409,16 @@ class iControlDriver(LBaaSBaseDriver):
                 provisioning_status == plugin_const.PENDING_UPDATE):
             if timed_out:
                 operating_status = (lb_const.OFFLINE)
+                if provisioning_status == plugin_const.PENDING_CREATE:
+                    loadbalancer['provisioning_status'] = \
+                        plugin_const.ERROR
+                else:
+                    loadbalancer['provisioning_status'] = \
+                        plugin_const.ACTIVE
             else:
                 operating_status = (lb_const.ONLINE)
-            loadbalancer['provisioning_status'] = plugin_const.ACTIVE
+                loadbalancer['provisioning_status'] = \
+                    plugin_const.ACTIVE
 
             self.plugin_rpc.update_loadbalancer_status(
                 loadbalancer['id'],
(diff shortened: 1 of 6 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
```diff
@@ -1381,19 +1399,25 @@ def _update_listener_status(self, service):
...
-        provisioning_status = loadbalancer['provisioning_status']
+        loadbalancer = service.get('loadbalancer', {})
+        provisioning_status = loadbalancer.get('provisioning_status',
+                                               plugin_const.ERROR)
 
         if (provisioning_status == plugin_const.PENDING_CREATE or
                 provisioning_status == plugin_const.PENDING_UPDATE):
-            operating_status = (lb_const.ONLINE)
+            if timed_out:
+                operating_status = (lb_const.OFFLINE)
```
mattgreene: If a member is timed_out, you set prov_status to ERROR.  Why not do the same here?  I think we've failed to fully provision the loadbalancer due to missing network configuration.

richbrowne: So, if we fail to provision a member, we know that it is in error.  For a loadbalancer, we need to conditionally set it to an error state if it is PENDING_CREATE.  Otherwise, we let neutron set the status to DEGRADED, i need to add a check here.  Good catch.

richbrowne:
Added the changes and retested.  Please squash before merging.

PR #610, opened 2017-03-03
add --debug to testenv operations run by buildbot

### LINKED ISSUE #609: testenv logs generated during builbot runs are ephemeral

zancas:
Issues:
Fixes #609

Problem: testenv logs generated by buildbot are lost
after the run

Analysis: by adding the debug flag we set testenv to
send logs to stdout where they will be caught by the
buildbot reporting tool

Tests: The output of the testenv commands will now show up in
nightly.

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -142,7 +142,7 @@ singlebigip:
 
 setup_singlebigip_tests: 
 	@echo executing $@
-	testenv create --name $${!DEVICEVERSION} \
+	testenv --verbose create --name $${DEVICEVERSION} \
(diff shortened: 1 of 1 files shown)


pjbreaux:
Do we intend to ever remove this? Could the output be too much to reasonably read through?

zancas:
It will increase the information in the log file.

pjbreaux:
The idea is that you run this manually and you don't see any issue that you suspect may occur in nightly? So you're putting this here to figure out problems when things actually go wrong?

PR #622, opened 2017-03-10
place unstable tests in the noregression manifest

### LINKED ISSUE #611: failing tests in the nightlies add noise and varaibles

zancas:
Issues:
Fixes #611

Problem: The build infrastructure is not yet stable.  To
stabilize it we need to discover the root causes of the
instability.  It's possible that poorly managed (torn down,
or setup) tests introduce instability into the regression
run.   It's likely that tests that have this effect will
fail.

Analysis: To isolate causes of instability in the regression
infrastructure we are marking unstable tests "no regression"
in manifest files that are external to the test code itself.

Tests: The determination of which tests are unstable and stable
was conducted using the bbot sandbox:
https://bldr-git.int.lineratesystems.com/kevin/bbot-sandbox

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -153,8 +153,9 @@ setup_singlebigip_tests:
 run_singlebigip_tests: 
 	@echo executing $@
 	tox -e singlebigip --sitepackage -- \
+	    --ignore test_network.py \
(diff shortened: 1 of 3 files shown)


zancas:
@pjbreaux @mattgreene I'm not sure if @mattgreene can merge this...  I think yes.

PR #646, opened 2017-04-04
Default pool ID ignored in listener update

### LINKED ISSUE #645: Default pool ID ignored in listener update
#### Description
Agent fails to set virtual server pool reference on BIG-IP in response to lbaas-listener-update --default-pool-id

#### Agent Version
9.2.0

#### Operating System
SuSE

#### OpenStack Release
Mitaka

#### Description
 We are seeing issues with the ‘detached’ pool functionality. Specifically, when we issue and update listener command from the CLI with a default pool set we see the listener default pool is set in Neutron, but in the VS on the devices the pool is set to None.
 
I think I see why, I just don’t understand how it can ever have worked, unless I’m missing something. As far as I can see everything on the agent is done with reference to a service dict, which is passed via RPC from the driver. Pools are built in the service builder as a list of dicts with key ‘pools’ https://github.com/F5Networks/f5-openstack-lbaasv2-driver/blob/mitaka/f5lbaasdriver/v2/bigip/service_builder.py#L133-L134
 
In the agent code I see this method  https://github.com/F5Networks/f5-openstack-agent/blob/master/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py#L127-L138 which as far as I can tell is never called  (https://github.com/F5Networks/f5-openstack-agent/search?utf8=%E2%9C%93&q=get_vip_default_pool&type=)
 
I’m also seeing this being used to get a pool from the service service[“pool”] e.g.  https://github.com/F5Networks/f5-openstack-agent/blob/master/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py#L95-L98  and it’s not clear to me from the driver service builder code how the key “pool” can exist in the service dict.
 
Can you confirm the lbaas-listener-update –default-pool xxxx yyyy works? It’s critical to our customers use case, specifically for blue/green deployment scenarios.

#### Deployment
Multiple BIG-IPs.

jlongstaf:
Issues:
Fixes #645

Problem: default_pool_id  is not processed by agent.

Analysis:

Tests:

@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
@@ -128,9 +128,6 @@ class LBaaSBuilder(object):
                 if pool:
                     svc['pool'] = pool
 
-            svc['pool'] = self.get_pool_by_id(
-                service, listener.get('default_pool_id', ''))
-
             if listener['provisioning_status'] == plugin_const.PENDING_UPDATE:
                 try:
                     self.listener_builder.update_listener(svc, bigips)
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
@@ -122,6 +122,15 @@ class LBaaSBuilder(object):
                    "listener": listener,
                    "networks": networks}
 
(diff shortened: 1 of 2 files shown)


jlongstaf:
I am closing this pull request because I don't believe this is really an issue with the agent code. I will update the original issue  [#645](https://github.com/F5Networks/f5-openstack-agent/issues/645) with my findings.

Issue #684, opened 2017-05-03
F5 agent not reporting error status back to Neutron


jlongstaf:
#### Agent Version
8.3.0

#### Operating System
SUSE

#### OpenStack Release
Liberty

#### Description
We have a customer report of scenario where creating LBaaS v2 load balancer shows ACTIVE status in OpenStack, but the F5 agent had errors configuring the BIG-IP and did not report the load balancer status as ERROR back to Neutron.

#### Deployment
Undercloud deployment, two BIG-IPs HA

PR #690, opened 2017-05-04
Fix KeyError exception when updating pool member

### LINKED ISSUE #680: Member updates failing due to missing address
Hi, 

We are seeing the following error when trying to update pools https://gist.github.com/abattye/29dffbadf1fbcdd88b293847f2a66393. It maybe related to https://github.com/F5Networks/f5-openstack-lbaasv2-driver/issues/544 because we are seeing this on pools which have corrupted nova managed port, however its may be unrelated. It appears the exception is not caught and the update to all pools related to the LB are rolled back. 

We have patched to just guard against the missing address, but it would be interesting to know why the member address is missing. This is the dict, which I think is returned from the API call to the device, {'session': 'user-enabled', 'ratio': 1, 'name': u'10.180.0.35%5:443', 'partition': u'Project_b7c88763cfb944349b6ac8b0c9850f15'}

We are working with the 9.3.0 codeline.

Thanks

Andrew

jlongstaf:
@pjbreaux 
Fixes #680 #684 

#### What's this change do?
1. Adds optional argument to member.pop() to prevent KeyError exception.
2. Adds exception handling for caller of update_member() in order to set status correctly.

pool_service.py, update_member()

#### Any background context?
member.pop() without optional param works fine for single BIG-IP installations. But, when deployed in multiple BIG-IP installations, the second call to pop() will fail with a KeyErrror. This fix prevents the error. Also, added exception handling to ensure that any other error that might be thrown is caught and status updated.

Test:
test_single_member_weight_updates.py::test_single_member_weight_updates PASSED

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
@@ -272,7 +272,11 @@ class LBaaSBuilder(object):
                         )
                         raise f5_ex.MemberCreationException(err.message)
                     else:
-                        self.pool_builder.update_member(svc, bigips)
+                        try:
+                            self.pool_builder.update_member(svc, bigips)
+                        except Exception as err:
+                            member['provisioning_status'] = plugin_const.ERROR
+                            raise f5_ex.MemberUpdateException(err.message)
                 except Exception as err:
                     member['provisioning_status'] = plugin_const.ERROR
                     raise f5_ex.MemberCreationException(err.message)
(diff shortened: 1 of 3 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
```diff
@@ -272,7 +272,11 @@ def _assure_members(self, service, all_subnet_hints):
                         )
                         raise f5_ex.MemberCreationException(err.message)
                     else:
-                        self.pool_builder.update_member(svc, bigips)
+                        try:
+                            self.pool_builder.update_member(svc, bigips)
+                        except Exception as err:
```
pjbreaux: Could we create a unit test to ensure that the exception is raised appropriately and the provisioning status is set correctly?

jlongstaf: Added new unit test in f5_openstack_agent/lbaasv2/drivers/bigip/test/test_lbaas_builder.py.

PR #716, opened 2017-05-15
Make setup failure cause all tests to not run.

### LINKED ISSUE #715: Makefile bug causes tests to run after a singlebigip-setup fail
#### Agent Version
8.3.1.b1

#### Description
This is a test infrastructure-only bug.

Because of an error in the Makefile logic a failure to setup "singlebigip" tests causes a single kind of the test run to be skipped, and then attempted to run subsequent tests.   The solution is to cause ALL tests to NOT run in the case of the setup target failing.

zancas:
Issues:
Fixes #715

Problem: A logic bug in the Makefile caused the automated tests to
attempt to run even though setup had failed.

Analysis:  The fix is to make 0 tests attempt a run upon setup fail.

Tests: The fix was tested by showing that a setup fail resulted in
0 attempted test runs in my sandbox environment.

@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -126,6 +126,7 @@ run_undercloud_tests:
 # This rule is run once per device type XX.Y.Z
 run_overcloud_tests:
 	@echo executing $@
+	$(MAKE) -C . setup_singlebigip_tests &&\
 	$(MAKE) -C . singlebigip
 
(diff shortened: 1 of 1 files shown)


pjbreaux:
And your success and failure case worked in your sandbox? @zancas

PR #717, opened 2017-05-15
Reactivate disconnected service tests

### LINKED ISSUE #647: Disconnected service tests not running in buildbot nightly
* Title: Disconnected service tests not running in buildbot nightly
* Details: This is only applicable to buildbot nightly runs. Changes should be localized to the systest directory from top-level.

#### Description
It appears that there's some level of indirection that is being used to both exclude certain tests and run others in the systest directory and makefile. For example, in [this](https://github.com/F5Networks/f5-openstack-agent/tree/liberty/systest/projects/disconnected_singlebigip_f5-openstack-agent_liberty-11_5_4-overcloud) location, there should be a file pointing pytest to where the actual test suite or test directory is located, but it's not there. I'll add that file.

zancas:
Issues:
Fixes #647

Problem: disconnected service tests were not running

Analysis: The makefile target had a stale path.

Tests: These tests were confirmed to pass in my local bbot sandbox.
I also verified that test results were written in the expected location.

### DIFF
--- a/systest/Makefile
+++ b/systest/Makefile
@@ -217,7 +217,7 @@ run_disconnected_service_tests:
 		--symbols $(MAKEFILE_DIR)/testenv_symbols/testenv_symbols.json \
 		--autolog-outputdir $${GUMBALLS_PROJECT} \
 		--autolog-session $${GUMBALLS_SESSION} \
-	    $${RUNTARGETS}
+	    ../test/functional/neutronless/disconnected_service/
 
 cleanup_singlebigip:
 	@echo executing $@


systest/Makefile
```diff
@@ -217,7 +217,7 @@ run_disconnected_service_tests:
 		--symbols $(MAKEFILE_DIR)/testenv_symbols/testenv_symbols.json \
 		--autolog-outputdir $${GUMBALLS_PROJECT} \
 		--autolog-session $${GUMBALLS_SESSION} \
-	    $${RUNTARGETS}
+	    ../test/functional/neutronless/disconnected_service/
```
pjbreaux: Are you migrating away from the symlink method in the singlebigip tests?

PR #739, opened 2017-05-25
Creating or changing loadbalancer admin state does not change virtual…

### LINKED ISSUE #704: Creating or changing loadbalancer admin state does not change virtual address enabled state
Title: Creating a Load Balancer with Admin State False should create a virtual address which is disabled and changing a loadbalancer admin state should alter the enable / disable status of the virtual address.

#### OpenStack Release
9.3

#### Agent Version
9.3.0

#### Operating System
Centos7

#### OpenStack Release
Mitaka

#### Description
If you create a load balancer with admin_state_up = False, the virtual address should be disabled. 

#### Deployment
From the neutron CLI issue:

neutron lbaas-loadbalancer-create --name test_lb01 [subnetid] --admin-state-down

When an lbaas-loadbalancer-update with --admin-state-up is issued, the virtual address should be enabled.

Once created making an REST call to change admin_state_up attribute to False does not disable the virtual address for the loadbalancer.

Once created making an REST call to change admin_state_up attribute to False does not enable the virtual address for the loadbalancer if it was disabled.

jlongstaf:
@pjbreaux 
Fixes  #704 

#### What's this change do?
Creating a Load Balancer with Admin State False should create a virtual
address which is disabled and changing a loadbalancer admin state should alter
the enable / disable status of the virtual address.

Added enabled attribue to VirtualAddress class, setting value
to yes/no based on service object attribute admin_state_up true/false.

virtual_address.py

#### Any background context?
No.

### DIFF
--- a/test/functional/neutronless/loadbalancer/test_loadbalancer_admin_state.py
+++ b/test/functional/neutronless/loadbalancer/test_loadbalancer_admin_state.py
@@ -55,6 +55,15 @@ def test_create_update_lb_state(bigip, services, icd_config, icontrol_driver):
     Update loadbalancer using admin-state-up and eval virtual address
     enabled value.
 
+    Note: this tests loadbalancer admin_state_up and how those values
+    are translated to the BIG-IP virtual address. The vip_port data in the
+    service object does not affect the virtual address enabled value,
+    only the loadbalancer admin_state_up will change the virtual address
+    enabled value.
+
+    Also, admin-state-down (no value allowed) is permitted only in create
+    commands. admin-state-up (with True/False values) is permitted only updtae
+    commands.
 
     :param bigip: BIG-IP under test
     :param services: list of service objects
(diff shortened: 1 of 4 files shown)


test/functional/neutronless/loadbalancer/test_loadbalancer_admin_state.py
```diff
@@ -0,0 +1,149 @@
...
+    service = service_iter.next()
+    lb_reader = LoadbalancerReader(service)
+    env_prefix = icd_config['environment_prefix']
+
+    folder = '%s_%s' % (env_prefix, lb_reader.tenant_id())
+
+    # Make sure we are starting clean.
+    assert not bigip.folder_exists(folder)
+
+    # Create loadbalancer, admin_state_down
```
pjbreaux: Could you clarify with each comment what the admin state of the lb and the port is?

jlongstaf: Sorry for the confusion. My comment for the first operation was incorrect (it should not have stated admin-state-down). I changed that comment and added additional comments to show 1) the Neutron CLI command, and 2) the service object loadbalancer admin_state_up value. Note that the admin_state of the vip_port data structure in the service object does not affect the virtual address object enabled attribute. Only loadbalancer admin_state does that. Hope this helps readers understand the test more clearly.

PR #745, opened 2017-05-26
Upgrade agent to use f5-sdk version 2.3.3

### LINKED ISSUE #744: Upgrade agent to use f5-sdk version 2.3.3
#### OpenStack Release
Liberty

#### Description
Need to upgrade agent to use 2.3.3. This is necessary to pick up pool member modify() fix which will fix agent issue setting pool member admin_update-up.

jlongstaf:
@pjbreaux 
Fixes #744 

#### What's this change do?
Requires that agent use version 2.3.3 of the f5-sdk.

setup.py

#### Any background context?
2.3.3 is necessary to fix an agent issue with setting admin_state_up attribute of pool members.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,7 +74,7 @@ version = VERSION
 release = VERSION
 
 # F5 SDK release version should be set here
-f5_sdk_version = '2.3.2'
+f5_sdk_version = '2.3.3'
 # F5 icontrol REST version should be set here
 f5_icontrol_version = '1.3.0'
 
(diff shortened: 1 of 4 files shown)


pjbreaux:
There are other changes to make here.

https://github.com/F5Networks/f5-openstack-agent/blob/liberty/setup.cfg
https://github.com/F5Networks/f5-openstack-agent/blob/liberty/f5-openstack-agent-dist/deb_dist/stdeb.cfg

pjbreaux:
And here as well. https://github.com/F5Networks/f5-openstack-agent/blob/master/docs/conf.py

We have a task to have a single place to update this version. It's in Jira.

ssorenso:
I did not know about the `docs/conf.py` aspect... The other ones are, indeed, required still for the agent.  The SDK's automation, if moved over, would reduce all of that to a `./setup_requirements.txt` (based upon how it's designed today).  This automation could handle other forms for the versioning to be passed through.  The issue, known, of passing the automation into the agent is that it would require two copies of the same code.

PR #746, opened 2017-05-30
Add ESD to a listener with a default pool with existing session persistence

### LINKED ISSUE #723: Add ESD to a listener with a default pool with existing session persistence does not implement lbaas_persist or lbaas_fallback_persist
#### Agent Version
9.3.0

#### Operating System
CentOS 7.3

#### OpenStack Release
Mitaka

#### Description
Adding ESD which has tags lbaas_persist and lbaas_fallback_persist to a listener which has a associated pool which already has session-persistence applied, does not apply the ESD tags properly.

#### Deployment
With an validated ESD called dmzmobile with:

```
"dmzmobile": {
   "lbaas_persist": "cookie",
    "lbaas_fallback_persist": "source_addr"
  }

```
```
neutron lbaas-loadbalancer-create --name test_lb01 dmz1 
neutron lbaas-listener-create --loadbalancer [lbid] --protocol TCP --protocol-port 443 --name test_lb01_listener
neutron lbaas-pool-create --listener [lnid] --session-persistence type=SOURCE_IP --name test_lb01_pool --lb-algorithm ROUND_ROBIN --protocol TCP 
neutron lbaas-l7policy-create --listener [lnid] --name dmzmobile --action REJECT 
```
Produces a virtual server with only source_add persistence applied. 

The source of truth of the configuration needs to be the ESD, not the pool or listener attributes.

jlongstaf:
@szakeri 
Fixes #723

#### What's this change do?
Problem: Persistence profiles applied by an ESD can be overwritten by
pool session persistence. ESDs should take precedence.

Analysis: Assure method operations have pools following l7policies, so
that pools with session persistence end up replacing persistence set
in L7 polcies ESDs. Change order so that L7 policies follow both
pool create and delete assure methods.

Tests: existing tests: tempest test_esd.py and agent test_esd.py.
Also, create new test, test_esd_pools.py.

lbaas_builder.py

#### Any background context?
Session persistence in LBaaS is a pool attribute. In BIG-IP/ESD it is a listener/virtual server attribute. Pool operations (create/delete) can set session persistence on a virtual server so we need to make sure the ESD operations happen after both pool create and delete.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
@@ -60,10 +60,6 @@ class LBaaSBuilder(object):
 
         self._assure_listeners_created(service)
 
-        self._assure_l7policies_created(service)
-
-        self._assure_l7rules_created(service)
-
         self._assure_pools_created(service)
 
         self._assure_monitors(service)
@@ -71,6 +67,10 @@ class LBaaSBuilder(object):
         self._assure_members(service, all_subnet_hints)
 
         self._assure_pools_deleted(service)
+
+        self._assure_l7policies_created(service)
+
+        self._assure_l7rules_created(service)
 
         self._assure_l7rules_deleted(service)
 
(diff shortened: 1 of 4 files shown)


zancas:
Hi @szakeri can you let me know when you've approved this?   It affects the handling of the failing tests in our "green core" which is my responsibility.

PR #753, opened 2017-06-02
Fix permissions issue where Jenkinsfile downloads systest results files.

### LINKED ISSUE #752: Fix permissions issue where Jenkinsfile downloads systest results files.
* Title: Fix permissions issue where Jenkinsfile downloads systest results files.

#### Description
- Nightly regression jobs report errors like "cp: preserving times for 'x': Operation not permitted" at the point where they try to download systest results logs.
- While not a blocker (because the job *does* run to completion and *does* successfully download the results files), this unexpected error is a nuisance because it causes the Jenkins weather report to remain stormy.

larkinkevin:
@zancas, @pjbreaux

Fixes #752 

#### What's this change do?
Fixes a bug in the Jenkinsfile.

n/a

#### Any background context?
no

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ pipeline {
                     # - copy results files to nfs
                     #   (note that the nfs results directory is mounted inside
                     #   the CI worker's home directory)
-                    cp -rp $WORKSPACE/systest/test_results/* ~/results/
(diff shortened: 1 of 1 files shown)


zancas:
I applied this change by pushing a patch, (so we didn't have to wait on Kevin to rebase).

PR #769, opened 2017-06-14
Route domain cache lookup returns wrong route domain

### LINKED ISSUE #768: Route domain cache lookup returns wrong route domain

#### Agent Version
9.3.1

#### Operating System
CentOS

#### OpenStack Release
Mitaka

#### Description
An incorrect route domain ID is returned from the route domain cache. There is an error in the function that searches the cache, so that when a cache entry is empty for a tenant, or the net name is not found in a cached entry, the function returns an incorrect route domain ID (the last route domain ID).

The function in error is get_route_domain_from_cache() in network_service.py. The function should return None if no positive match of net_short_name is found in the rds_cache.

#### Deployment
Two BIG-IPs in HA pair, version 12.1.2

jlongstaf:
@pjbreaux 
Fixes #768

#### What's this change do?
An incorrect route domain ID can be returned from the route domain cache. There is an error in the function that searches the cache, so that when a cache entry is empty for a tenant, or the net name is not found in a cached entry, the function returns an incorrect route domain ID (the last route domain ID). Modified the function get_route_domain_from_cache() to raise an exception when no entry is found, then handle the exception in the calling function.

No unit tests currently cover testing get_route_domain(). Created a new unit test that tests this change, and also tests that the change didn't adversely affect assign_route_domain() which is the only caller of get_route_domain().

Changes to get_route_domain() in network_service.py

#### Any background context?
The code as originally implemented functioned correctly. A code change in a Feb 14 commit introduced this bug.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_network_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_network_service.py
@@ -1,0 +1,223 @@
+# coding=utf-8
+# Copyright 2017 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from collections import namedtuple
+from f5_openstack_agent.lbaasv2.drivers.bigip import exceptions as f5_ex
+from f5_openstack_agent.lbaasv2.drivers.bigip.network_service import \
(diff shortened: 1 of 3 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/test/test_network_service.py
```diff
@@ -0,0 +1,223 @@
...
+            network_service.get_route_domain_from_cache(network)
+        assert 'vlan-606' in str(excinfo.value)
+
+        # invalid network data
+        with pytest.raises(f5_ex.InvalidNetworkType):
+            network['provider:segmentation_id'] = 604
+            network['provider:network_type'] = ''
+            network_service.get_route_domain_from_cache(network)
+
+    def test_assign_route_domain(self, network_service, network, subnet):
```
pjbreaux: An additional test that is needed is to show that we log the message we expect to when we call assign_route_domain and we get a RouteDomainCacheMiss. The other test is just to ensure we actually want to raise f5_ex.InvalidNetworkType up when we call assign_route_domain and a bad network type or seg id has been given. That would raise in get_neutron_net_short_name, then go up to get_route_domain_from_cache, and finally up to assign_route_domain.

PR #793, opened 2017-07-03
Pool will not be recreated if neutron-lbaas show it as 'ACTIVE' and pool does not exist on device

### LINKED ISSUE #783: Pool will not be recreated if neutron-lbaas show it as 'ACTIVE' and pool does not exist on device
#### OpenStack Release
liberty, but should be merged forward

#### Description
When a pool is in an 'ACTIVE' or 'ERROR' state in neutron-lbaas, and the pool somehow disappears from the BIG-IP device, the _assure_pools_created method in lbaas_builder.py does not properly recreate the pool because of faulty if/then logic. In fact, if 'ACTIVE' or 'ERROR' is seen as the provision status, we will try to update the pool instead of creating it. We should change this logic to better fit the listener method.

We should also ensure that other objects do not suffer from this same faulty logic. I believe all objects can be treated the same way.

Note: this was initially found in a customer deployment where the BIG-IP device was restarted (without save /sys config) and the pool disappeared upon reboot. Then the agent did not recreate the pool properly when the device came back online.

pjbreaux:
@jlongstaf 
Fixes #783 

#### What's this change do?
Corrected the error and created unit tests to test the new
functionality.

#### Any background context?
When a pool is in an 'ACTIVE' or 'ERROR' state in neutron-lbaas, and the
pool somehow disappears from the BIG-IP device, the
_assure_pools_created method in lbaas_builder.py does not properly
recreate the pool because of faulty if/then logic. In fact, if 'ACTIVE'
or 'ERROR' is seen as the provision status, we will try to update the
pool instead of creating it. We should change this logic to better fit
the listener method.

The new units tests pass. We will create a new tempest API test in the
driver to check this functionality as well.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
@@ -150,8 +150,8 @@ class LBaaSBuilder(object):
 
                 try:
                     # create or update pool
-                    if pool['provisioning_status'] == \
-                            plugin_const.PENDING_CREATE:
+                    if pool['provisioning_status'] \
+                            != plugin_const.PENDING_UPDATE:
                         self.pool_builder.create_pool(svc, bigips)
                     else:
                         self.pool_builder.update_pool(svc, bigips)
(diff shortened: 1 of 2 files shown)


pjbreaux:
@jlongstaf: just rebased and pushed.

PR #810, opened 2017-07-24
bump version for release

szakeri:
@pjbreaux


ssorenso:
Regardless as to the Travis failure, I have not looked at this...

This has to be done in a few places:
* [stdeb.cfg](https://github.com/F5Networks/f5-openstack-agent/blob/master/f5-openstack-agent-dist/deb_dist/stdeb.cfg)
* [setup.cfg](https://github.com/F5Networks/f5-openstack-agent/blob/master/setup.cfg)

This is still needed until the universal truth PR is approved.... for today, I would suggest manually changing these files.

PR #812, opened 2017-07-25
Feature.liberty.universal true 765

### LINKED ISSUE #765: Create Universal Source of Truth for Dependencies
* Title: Create Universal Source of Truth for Dependencies
* Attachments: None
* Details:
  * Create a means by which to reference in just one place where all dependencies are to execute `pip install ./` for the `f5-openstack-agent` lies
  * This does not effect or impact documentation sources that will still need to be edited manually for each release

#### OpenStack Release
TBD - First to be checked into Liberty

#### Description
This enhancement should be like the [f5-sdk's RPM and DEB package generation scripts](https://github.com/F5Networks/f5-common-python/issues/980); however, the current scope is to only create the one file as a universal truth.

ssorenso:
@jlongstaf 
Fixes #765

#### What's this change do?
* Limitedly introduces universal truth solution from f5-common-python
  * Introduces the [construct_setups.py](https://github.com/F5Networks/f5-common-python/blob/development/f5-sdk-dist/scripts/construct_setups.py)
  * Introduces the [configure.py](https://github.com/F5Networks/f5-common-python/blob/development/f5-sdk-dist/scripts/configure.py)
* Only uses what is needed in the above libraries
* Introduces a new handler that simply performs the universal truth aspect from f5-common-python

./f5-openstack-agent-dist/

#### Any background context?
* This code exists in f5-common-python; however, rather than duplicate the entire effort, this seeks an interim of simply porting over the same functionality for just this specific task
* This full coded feature will later be either outsourced or a new github repo created to house it


pjbreaux:
@ssorenso: you've got a failing build. I'd like to discuss this task with you next week. Thanks.

pjbreaux:

pjbreaux:
@ssorenso: I'm testing build on PR for the agent currently, thus the message from me above. I will create a new email and github user this week to change this over to a user that is not me! Nothing to do currently. Thanks!

Issue #803, opened 2017-07-14
f5_common_networks .ini file setting influence what is in the Common partition


ssorenso:
* Title: f5_common_networks .ini file setting influence what is in the Common partition
* Attachments: None
* Details: 

#### OpenStack Release
Mitaka onward

#### Description
This enhancement will handle the inclusion of this setting's influence on whether an L2ServiceBuilder object is placed into the Common partition or not.  This is later to allow for RBAC functionality.

#### Agent Version
TBD

#### Operating System
All

#### OpenStack Release
Mitaka onward

PR #813, opened 2017-07-25
Feature.mitaka.common networks functional tests 807

### LINKED ISSUE #807: Common Networks ini setting having functional tests
* Title: Common Networks ini setting having functional tests
* Attachments:None
* Details: previously committed #803 change did not include functional tests due to the fact that this work was divided across sprints.

#### OpenStack Release
Mitaka

#### Description
This is a task to handle the creation of functional tests for the ini setting.  This should include a lbaas-driver systest test and a neutronless test.

#### Agent Version
TBD

#### Operating System
All

#### OpenStack Release
Mitaka onward

#### Description
To offer full test coverage, the unit test aspect of this feature was committed within #803; however, there needs to be functional tests that test the integration of these changes within the system as a whole.  That said, this test will be partially testing features that are not yet integrated completely within the design.  Aspects would include further division of objects that show up in the Common partition versus the tenant partition.  The missing object transfers, thus far, are Route Domains and SNAT Pools.  These should both be on the Common partition when this setting is enabled.  Later tasks will handle the transfer of these objects feature-wise.

The tests that are expected to fail will be put into a pytest's skipped setting mode so as not to impact nightly's numbers passed their presence.

ssorenso:
@jlongstaf 
Fixes #807 
WIP #807 - neutron testing infrastructure add: Keeping environment clean via a new decorator object
WIP #807 - `test_snat_common_network.py` now cleans up after itself using less intrusive means

#### What's this change do?
* 32587ae89ff170abf0e9f030ad8822d864b44561
  * Makes `test_snat_common_network.py` clean up after itself more appropriately
* dbc003b45a885e5c46434771fb0f26037d6ad4f7
  * **Tests will run without this; however, as a group, it will fail without 32587ae89ff170abf0e9f030ad8822d864b44561**
  * Creates a test that will validate the setup of a common network via the snat exchange up to a member
  * Creates a test that will test that Route Domains are initiated in the Common partition and that the SNATPool will also; however, this has not been implemented yet
    * This test is skipped for now via `pytest.mark.skip("skip", reason=...)`
* e710b2317bd107d7d51479b9320617dd0150826d
  * Creates a new neutronless BIG-IP handling object, `Resource4TestTracker`
    * This object, using the `with` block in python, will handle any cleanup from failed tests

* e710b2317bd107d7d51479b9320617dd0150826d
  * or `./neutronless/conftest.py`

#### Any background context?
This is to satisfy the agent functional testing for the `f5_openstack_agent.ini`'s `f5_common_networks` setting addition performed earlier by #803

### DIFF
--- a/test/functional/neutronless/conftest.py
+++ b/test/functional/neutronless/conftest.py
@@ -13,16 +13,216 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import json
+import os
+import pytest
+import re
+import traceback
+# import sys
+
+from inspect import currentframe as cf
+from inspect import getframeinfo as gfi
+from inspect import getouterframes as gof
+from collections import deque
+from collections import namedtuple
+from copy import deepcopy
 
 from .testlib.bigip_client import BigIpClient
 from .testlib.fake_rpc import FakeRPCPlugin
 from f5_openstack_agent.lbaasv2.drivers.bigip.icontrol_driver import \
     iControlDriver
 
-from copy import deepcopy
-import json
-import os
-import pytest
+DEBUG = True  # sys.flags.debug
+
+
+class Resource4TestTracker(object):
+    """Creates an object meant for tracking states assigned to the BIG-IP
+
+    This testing object will track the states that are assigned by the test
+    from inside of the service operations on the agent.  Upon closure, it will
+    destroy the objects in the revere order of the states that were assigned.
+
+    Additionally, it will take the service object and attempt the cleanup of
+    the provided cleanup# order.  Thus, if there is a set of keys:
+        ['cleanup0', 'cleanup10', 'cleanup2']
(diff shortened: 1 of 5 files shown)


test/functional/neutronless/conftest.py
```diff
@@ -39,6 +239,14 @@ def fin():
     return bigip
 
 
+def debug_msg(status):
```
ssorenso: Note that this is quite handy... upon a failed test, it will output the debug statement into the STDOUT with the line from the test that called it; thus, you know when/where the deployed state was performed.

pjbreaux:

pjbreaux:
@ssorenso: I'm testing build on PR for the agent currently, thus the message from me above. I will create a new email and github user this week to change this over to a user that is not me! Nothing to do currently. Thanks!

ssorenso:
Fetched && Rebased

ssorenso:
test this please

PR #828, opened 2017-08-01
Pool does not associate with listener if pool is created first

### LINKED ISSUE #819: Pool does not associate with listener if pool is created first
#### OpenStack Release
mitaka

#### Description
When a pool is created and then the listener is created, the pool should be linked to the listener via the _assure_pools_created method in lbaas_builder.py. This does not work because the pool already exists on the device and thus raises a 409 exception up to the function above in lbaas_builder.py, causing the function to halt there. The listener is never associated with the pool then. The offending line is here:

https://github.com/F5Networks/f5-openstack-agent/blob/mitaka/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py#L164

We should wrap the lines with a check for the 409, and continue if so, otherwise, raise.

#### Agent Version
<Fill in the version you have installed, such as 2.0.1>

#### Operating System
<Fill in the host OS of the machine running the agent, such as Centos 7>

#### OpenStack Release
<Fill in the OpenStack release, such as Liberty>

#### Description
<Describe the bug in detail, steps taken prior to encountering the issue, yand a short explanation of you have deployed openstack and F5® agent>

#### Deployment
<Explain in reasonable detail your OpenStack deployment, the F5® OpenStack agent, and BIG-IP®(s)>
<Example: Single OpenStack controller with one F5® agent managing a cluster of 4 BIG-IP® VEs>
<Example: Three OpenStack controllers in HA, each with one standalone F5® agent managing a single BIG-IP® appliance>

pjbreaux:
@jlongstaf 

**Please don't merge until I've run the system tests**

Fixes #819 

#### What's this change do?
Modified the _assure_pools_created call to only raise an exception if an
HTTP error seen other than a 409. If a 409 is seen, the call proceeds
and it is assumed the pool exists.

#### Any background context?
When a pool is created and then the listener is created, the pool should
be linked to the listener via the _assure_pools_created method in
lbaas_builder.py. This does not work because the pool already exists on
the device and thus raises a 409 exception up to the function above in
lbaas_builder.py, causing the function to halt there. The listener is
never associated with the pool then.

Wrote new unit tests for this change. In addition to the system tests
that are currently failing in the driver, that should be enough to prove
this change is good.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
@@ -159,11 +159,15 @@ class LBaaSBuilder(object):
 
                 try:
                     # create or update pool
-                    if pool['provisioning_status'] \
-                            != plugin_const.PENDING_UPDATE:
-                        self.pool_builder.create_pool(svc, bigips)
-                    else:
-                        self.pool_builder.update_pool(svc, bigips)
+                    try:
+                        if pool['provisioning_status'] \
+                                != plugin_const.PENDING_UPDATE:
+                            self.pool_builder.create_pool(svc, bigips)
+                        else:
+                            self.pool_builder.update_pool(svc, bigips)
+                    except HTTPError as err:
+                        if err.response.status_code != 409:
+                            raise f5_ex.PoolCreationException(err.message)
 
                     # assign pool name to virtual
                     pool_name = self.service_adapter.init_pool_name(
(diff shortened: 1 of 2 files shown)


pjbreaux:
@szakeri: This is ready, as I have run the corresponding tests in the driver and they pass:

```
api/test_pools.py::PoolTestJSON::test_create_shared_detached_pools PASSED
api/test_pools.py::PoolTestJSON::test_create_shared_pools PASSED
```

pjbreaux:
@szakeri: also, please ignore the still running smoke test on PR. That was a test in my jenkins instance. The real build on PR completed and it turned green in the line that reads 'default -- Build finished'

PR #861, opened 2017-08-17
smoke tests on PR do not fail when a test setup failure occurs

### LINKED ISSUE #860: smoke tests on PR do not fail when a test setup failure occurs
#### OpenStack Release
<The earliest release in which you would like to see the enhancement>

#### Description
The build on PR smoke tests, which are run when a PR is opened in the agent, fail as expected when a test fails (tested in liberty), but the build does not fail when the tests error out.

pjbreaux:
@szakeri 
Fixes #860 

#### Any background context?
The build on PR smoke tests, which are run when a PR is opened in the
agent, fail as expected when a test fails (tested in liberty), but the
build does not fail when the tests error out.

Removed the dash before the setup target in the Makefile.


pjbreaux:
retest this please

PR #882, opened 2017-08-28
F5 network mitaka issue 730 merge

### LINKED ISSUE #730: agent should report as down to neutron when communication is lost to any associated BIG-IP
#### Agent Version
9.3.1b

#### Operating System
Centos 7.3

#### OpenStack Release
Mitaka

#### Description

#### Deployment

    `bigstart stop`

Provision lbaas objects.

jgruber:
@richbrowne 
Fixes #730 
...

#### What's this change do?
BIG-IPs now have independent status which reports to Neutron.
BIG-IPs can now go up and down with reconnect support.
BIG-IPs for which connections are reestablished can catch up their configurations.
BIG-IPs for which connections are reestablished have orphaned object support.
BIG-IPs now save their configurations.

Drinking

#### Any background context?
code changes testing on customer POC.


pjbreaux:

PR #923, opened 2017-09-14
WIP: Troubleshooting commit for troubleshooting PKG's

ssorenso:
Issues:
General packaging issues

Problem:
* Travis is given the same scripting as what was used locally
* Travis results in weird dependencies whereas locally works

Analysis:
This adds troubleshooting steps for viewing in Travis results

Test:
None, this is a troubleshooting step

@richbrowne 
WIP


ssorenso:
This is not the right fix, and was used to troubleshoot Travis... closing PR

PR #931, opened 2017-09-18
Travis change for greater tracking of RPM issues

### LINKED ISSUE #930: Travis change for greater tracking of RPM issues
* Title: Travis change for greater tracking of RPM issues
* Attachments: 
* Details: 

Travis passes when the rpm tests fail.  This is due to the fact that the rpm tests are not in the `scripts` section.  Instead, they are in the post-processing travis section, which is not MC for a pass/fail in travis.

#### OpenStack Release
All supported (liberty, mitaka, newton)

#### Agent Version
newton - 1e01e9688c3ccc0d67e96289e80dfa53a38b5ae3
mitaka - 1e01e9688c3ccc0d67e96289e80dfa53a38b5ae3
liberty - 7a9eb2d5ae9f6332a95b9589a89f3b324dc186ef

#### Operating System
Redhat CentOS 7 is the greatest impact

ssorenso:
Issues:
Fixes #930

Problem:
* RPM packages are getting created in Travis
  * With a rate of about 20%, random parsing errors can occur with deps

Analysis:
* The test protion of travis scripting was moved to 'script' section
  * This is a mission-critical section for declaring a test failure
  * Upon test failure, work around is to rebuild the test to get rpm

Tests:
* This is test code

@richbrowne 
#### What's this change do?
Moves the testing portion of Travis's testing code for RPM package building (and debian by proxy) into the Travis's `script` section.  This should cause the tests to fail if/when a package is not properly built.

.travis.yml file

#### Any background context?
This is only to address our tests failing if/when a travis build fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,17 +21,21 @@ install:
 script:
     - tox -e style,unit
     - python f5-openstack-agent-dist/scripts/universal_truth.py
+    - echo "setup.cfg:"
+    - cat setup.cfg
+    - echo "stdeb.cfg:"
+    - cat f5-openstack-agent-dist/deb_dist/stdeb.cfg
     - f5-openstack-agent-dist/scripts/package_agent.sh "redhat" "7"
     - f5-openstack-agent-dist/scripts/package_agent.sh "redhat" "6"
     - f5-openstack-agent-dist/scripts/package_agent.sh "ubuntu" "14.04"
     - sudo chown -R travis:travis ${DIST_REPO}/rpms/build
     - sudo chown -R travis:travis ${DIST_REPO}/deb_dist/*.deb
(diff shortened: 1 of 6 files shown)


richbrowne:
The rest of the commit looks fine.  Though I don't know why the packaging failed.

ssorenso:
So there's generally no reason to keep the `setup.cfg` and the `stdeb.cfg` files around even as templates... if later we add them to gain some additional settings, then that's fine, the existing handling will use them; however, I've set up defaults for these files to be used if they're not present.

ssorenso:
@richbrowne ready for re-review.  Please see latest commit for changes.  Put it in a separate commit to isolate non-critical updates.

PR #942, opened 2017-09-20
Issue.730

### LINKED ISSUE #730: agent should report as down to neutron when communication is lost to any associated BIG-IP

jgruber:
@richbrowne 
Fixes #730

#### What's this change do?
There was an issue in the 'create or update' workflow for virtuals on the BIG-IP because of the order of attribute processing compared to the pre-issue 730 code. The previous code looked at the Neutron provisioning_state of the Neutron listener to decide if it should try to create or update the BIG-IP virtual object. This won't work on hosts which are not in sync with provisioning. (why we changed things)

This change makes the create_listener method in listener_service make a call to the BIG-IP to see if the virtual exists and uses that to determine if a create or update is required. If the virtual service on the particular BIG-IP needs to be created.. it is created.. if it needs to be updated.. the exact same workflow that was in place before the changes is used (self.update_listener but for only the specific bigip).

Look at changes in lbaas_builder and listener_service.


pjbreaux:

pjbreaux:

richbrowne:
This was handled by PR #1085

PR #947, opened 2017-09-22
Test dependencies do not match community or driver specs.

### LINKED ISSUE #946: Dependency versions used in the smoke (and functional and unit) tests skewed from comnunity and driver tests
Several of the dependencies installed by using the `pip install -r requirements.functest.txt` command are different versions than those specified in the relevant constraints file:

https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=mitaka-eol

The resulting package versions also differed from those used by our driver tests (which match the above constraints).

While I have not yet checked the state of the `stable/newton` code I suspect a similar issue exists.

zancas:
@pjbreaux 
Fixes #946

#### What's this change do?
  By altering the way in which smoke tests are run I fix the test dependency version-skew issue.

The following smoke tests results show smoke tests passing using the container based approach.  This is a significant step towards migrating the agent to use testrunner containers.   I implemented this solution now, because the tasks of recording, aggregating, and publishing coverage results are also best carried out by our containerized test infrastructure..

The solution I implemented guarantees that the `agent` and `driver` tests have the exact same dependencies because their respective testrunners are built on the same image (I call it a layer).
So, it's not just the same constraints file and installation command, it's actually the same image.

Examine these test results:

http://10.190.25.149:49001/blue/organizations/jenkins/devs%2Fzancas_smoke/detail/zancas_smoke/25/pipeline


zancas:
Attached is an image showing some of the packages that are skewed:

delta

zancas:
I expect to discuss ramifications and put the final touches on this PR (squashing commits, pointing the Jenkinsfile "agent" at a production image, etc.) after review.

.gitignore
```diff
@@ -56,6 +56,7 @@ __pycache__/
 
 # Distribution / packaging
 .Python
+.contain
```
zancas: This is a convenience for me.

Jenkinsfile
```diff
@@ -5,7 +5,7 @@ pipeline {
         docker {
             label "docker"
             registryUrl "https://docker-registry.pdbld.f5net.com"
-            image "bdo/jenkins-worker-ubuntu-16.04:master"
+            image "openstack-test-agenttestrunner-zancas/mitaka:latest"
```
zancas: This should be updated to "prod" after all other issues are resolved.

Jenkinsfile
```diff
@@ -22,6 +22,7 @@ pipeline {
             steps {
                 sh '''
                     # - initialize env vars
+                    export JOB_BASE_NAME=12.1.2-overcloud_smoke
```
zancas: This hack comes out at the same time we update to "prod" above.

requirements.functest.txt
```diff
@@ -13,7 +13,6 @@ git+https://github.com/F5Networks/pytest-symbols.git
 mock==2.0.0
 pytest>=3.0.5,<4
 pytest-cov>=2.4.0,<3
-responses==0.5.1
```
zancas: By inspecting with "grep -r" I formed the belief that this package is never used.

requirements.unittest.txt
```diff
@@ -10,7 +10,6 @@ git+https://github.com/F5Networks/f5-openstack-lbaasv2-driver.git@mitaka
 mock==2.0.0
 pytest==3.0.3
 pytest-cov==2.3.1
-responses==0.5.1
```
zancas: See above.

systest/Makefile
```diff
@@ -93,8 +93,8 @@ TESTENVLOGDIR := $(RESULTS_DIR)/f5-openstack-agent_$(SANITIZED_BRANCH)/testenv
 #    Tests use tox for python variants.
 setup:
 	@echo executing $@
-	sudo -E -H $(SCRIPT_DIR)/install_test_infra.sh ;\
-	mkdir -p $(TESTENVLOGDIR)
```
zancas: Dependencies are now rolled into the `agenttestrunner` I a full test run will be necessary to land the above... but significantly the smoke test does pass.

systest/Makefile
```diff
@@ -93,8 +93,8 @@ TESTENVLOGDIR := $(RESULTS_DIR)/f5-openstack-agent_$(SANITIZED_BRANCH)/testenv
...
+	mkdir -p $(TESTENVLOGDIR) && \
+	sudo -E pip install -e $(PROJDIR)
```
zancas: This guarantees that we're actually testing against the code we Jenkins checked out for us!

systest/Makefile
```diff
@@ -253,15 +253,14 @@ run_disconnected_service_tests:
 
 run_smoke_tests:
 	@echo executing $@
-	tox -e singlebigip --sitepackage -- \
```
zancas: No need for tox.

zancas:
Here're the latest (most current test results):

http://10.190.25.149:49001/view/all/job/devs/job/za_smoke_two/3/console

zancas:
Bummer...   testenv fail...   possibly fighting with nightly at this point...

PR #948, opened 2017-09-22
added fixes to fix orphaned listeners

### LINKED ISSUE #730: agent should report as down to neutron when communication is lost to any associated BIG-IP

jgruber:
@richbrowne 
@jlongstaf 
Fixes #730 

#### What's this change do?
Added a fix where an AttributeError exception happens when no pool object is defined and added the orphaned object support for listeners when a loadbalancer is still ACTIVE, but the listener was deleted while a BIG-IP was disable (did handle that case previously).


pjbreaux:

richbrowne:
I am submitting this in another PR.

PR #949, opened 2017-09-22
Sync and check svc objects in ERROR status

### LINKED ISSUE #878: Common networking tweaks for full feature
* Title: Common networking tweaks for full feature
* Attachments:
* Details: For enhancement

#### OpenStack Release
Mitaka+

#### Description
This is to provide both feature and tests as an enhancement for common networks feature changes.

#### Agent Version
mitaka+

#### Operating System
Any supported

#### OpenStack Release
Mitaka+

#### Deployment
Not specific to deployment outside of common-partition based networking

ssorenso:
Sync and check svc objects in ERROR status

Issues:
Fixes #878

Problem:
* We see objects go into ERROR status, but never do anything further
* We also do not know a blanketing condition of ERROR for lower objects
* Objects concerned within the service object from lbaas

Analysis:
* This change will check for ERROR status in the service object
* It will then track the error status better and more appropriately

Tests:
Items have been moved from test_listener_services.py to be more useful
across multiple tests by moving the svc construction fixtures into
conftest, then inherited by the Test*Constructor objects.
Secondly, some minor house keeping was performed to isolate what was
needed here into the Constructor, Builder, and Test objects to follow a
constructor/builder/user pattern that is unique per test method.

@jlongstaf 
WIP #878 

#### What's this change do?
Enforces the svc[objs][ii] objects into an ACTIVE state after creation|update.  Then adds a check for svc-wide ERROR status and propogates it forward to the LB in the agent_manager.  By performing this, timeouts are handled more efficiently.

agent_manager.py

#### Any background context?
This is part of the common networking push.  Thus, the deviated branch.

it is still expected that
functional.neutronless.test_f5_common_networks
Will fail in its one test that is not skipped in smoke tests.  This is due to the fact that the snat pool is not deleted when the neutronless test cleans up after itself with an empty service object.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
@@ -569,9 +569,10 @@ class LbaasAgentManager(periodic_task.PeriodicTasks):  # b --> B
                 lb_id
             )
             self.cache.put(service, self.agent_host)
-            if not self.lbdriver.service_exists(service):
-                LOG.error('active loadbalancer %s is not on BIG-IP...syncing'
-                          % lb_id)
+            if not self.lbdriver.service_exists(service) or \
+                    self.has_provisioning_status_of_error(service):
+                LOG.info("active loadbalancer '{}' is not on BIG-IP"
+                         " or has error state...syncing".format(lb_id))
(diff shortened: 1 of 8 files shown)


ssorenso:
retest this please

PR #951, opened 2017-09-22
add constraints to requirements files

zancas:
@ssorenso 

Fixes: #946 

Please see the description in the issue and the comments in the requirements files.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/test/test_lbaas_builder.py
@@ -811,7 +811,7 @@ class TestLbaasBuilder(object):
         builder.get_l7policy_for_rule.assert_called_once_with(
             mock_service['l7policies'], mock_service['l7policy_rules'][0])
         l7policy.get.assert_called_once_with('name', None)
-        builder.driver.get_config_bigips.assert_called_once()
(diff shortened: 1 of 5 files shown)


zancas:
I was logged in as `f5openstackdevautobot`, so those comments are mine.

pjbreaux:
Once you are ready for review, please add a full commit message and fill out the full pull request template.

zancas:
constraints_ordering_expt


The uploaded screenshot shows how the behavior differs if we put the `-c constraints.txt` line as the last line in the requirements file.   In short, by putting that constraints directive at the end we cause an error in the case of a conflict, and the installation does not occur.  I believe that this is the correct behavior.

f5_openstack_agent/lbaasv2/drivers/bigip/test/test_lbaas_builder.py
```diff
@@ -811,7 +811,7 @@ def l7_check_being_esd(self, action_mock, target_method, check_log=False,
         builder.get_l7policy_for_rule.assert_called_once_with(
             mock_service['l7policies'], mock_service['l7policy_rules'][0])
         l7policy.get.assert_called_once_with('name', None)
-        builder.driver.get_config_bigips.assert_called_once()
```
zancas: This call is incompatible with the mitaka-eol constraints file.

Issue #963, opened 2017-09-26
Refactor fdb updates


jlongstaf:
#### OpenStack Release
Liberty

#### Description
Need to refactor how fdb entries are updated on the BIG-IP. Current implementation updates each individual fdb entry which does not scale well. For installations with a large number of VTEPs, performance can be very poor (> 50 seconds to update all records).

PR #964, opened 2017-09-26
Refactor fdb updates

jlongstaf:
@richbrowne 
DO NOT MERGE

#963 Refactor fdb updates

#### What's this change do?
Changes how fdb add/deletes are done in post_service_network function. Instead of updating each fdb entry indvidually, collects fdb records for all VTEPs (both loadbalancer and member), then updates each tunnel.  This reduces the number of BIG-IP operations from number of VTEPs to number of tunnels.

network_service.py

#### Any background context?
Number of VTEPs for a single loadbalancer can be in the hundreds, which is the likely performance bottleneck.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/l2_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/l2_service.py
@@ -763,12 +763,13 @@ class L2ServiceBuilder(object):
                 fdbs[tunnel_name]['records'] = dict()
 
             records = fdbs[tunnel_name]['records']
-            mac_addr = member['port']['mac_address']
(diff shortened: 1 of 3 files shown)


pjbreaux:
@jlongstaf: I'm going to kick this job to test that the agent build on PR is working. This is just a test of that functionality.

pjbreaux:
ok to test

pjbreaux:
@jlongstaf : the config on the smoke tests was wiped when the job was disabled. That is why you see two lines showing up referring to smoke tests above. If we open new pull requests, we should only get one message: 'Running smoke tests on PR'

jlongstaf:
Retest this please

PR #971, opened 2017-09-28
fix path name for test runs

zancas:
@ssorenso 

Fixes: #972

I believe this is the reason (or reasons) the nightly run failed.

When developing I was only testing against a single test run "run_neutronless_loadbalancer".  I forgot to modify the paths for the other test runs when I reverted from dev code to prod code.   DRYer code would make this kind of error less likely.

Additionally I note that the complete set of parameters passed to the test runner is _NOT_ specified in the Jenkinsfile.

This commit fixes two errors:
   1) each py.test invocation was overwriting .coverage from the previous invocation.
   2) py.test paths were incorrect

I also remove unused "coveragerc_bak" files.

### DIFF
--- a/.coveragerc_bak
+++ b/.coveragerc_bak
@@ -1,9 +1,0 @@
-[run]
-source = 
-    */lbaasv2/*
-
-[paths]
-source =
-    f5_openstack_agent/
-    /usr/lib/python2.7/site-packages/f5_openstack_agent/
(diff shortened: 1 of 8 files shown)


zancas:
@pjbreaux I still see 2 differnt references to smoke in this PR.   Is that the desired state?

zancas:
If I can tell if "review" is one of the "checks" listed above.

PR #1023, opened 2017-10-09
Fixes #1020 -- docs deployment

### LINKED ISSUE #1020: Docs did not publish for some tagged releases
#### Agent Version
10.0.0
10.1.0
9.4.0
8.3.6

#### OpenStack Release
stable/newton

#### Description
The documentation did not deploy for the tagged releases. 

- [ ] Deploy tagged versions by creating a new branch from each release tag.
- [ ] Troubleshoot and fix the deploy script.

jputrino:
@dflanigan 
@ssorenso 

Fixes #1020 

#### What's this change do?
Updates how the variable that determines the publication path for docs is set.
This should prevent us from having further issues with deployment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
     - PKG_RELEASE=$(python ${DIST_REPO}/scripts/get-version-release.py --release)
     - PKG_RELEASE_EL7=${DIST_REPO}/rpms/build/f5-openstack-agent-${PKG_VERSION}-${PKG_RELEASE}.el7.noarch.rpm
     - PKG_RELEASE_1404=${DIST_REPO}/deb_dist/python-f5-openstack-agent_${PKG_VERSION}-${PKG_RELEASE}_1404_all.deb
-    - RELEASE_TAG="v$(echo $PKG_VERSION | cut -d . -f 1,2)"
+    - RELEASE_TAG=v$(echo $PKG_VERSION | cut -d . -f 1,2)
 services:
 - docker
 language: python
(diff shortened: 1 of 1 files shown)


pjbreaux:

.travis.yml
```diff
@@ -6,6 +6,7 @@ env:
     - PKG_RELEASE=$(python ${DIST_REPO}/scripts/get-version-release.py --release)
     - PKG_RELEASE_EL7=${DIST_REPO}/rpms/build/f5-openstack-agent-${PKG_VERSION}-${PKG_RELEASE}.el7.noarch.rpm
     - PKG_RELEASE_1404=${DIST_REPO}/deb_dist/python-f5-openstack-agent_${PKG_VERSION}-${PKG_RELEASE}_1404_all.deb
+    - RELEASE_TAG="v$(echo $PKG_VERSION | cut -d . -f 1,2)"
```
ssorenso: Understanding that this is a 'picky' semantic change, I apologize at the level of 'meaning' at runtime; however, there's no reason to use the `"`.  Compile this with the fact that we're not using `"` anywhere else in these definitions without a clear reason, and I'm left with the thought that we should continue to use the same pattern and only use quotes when they're necessary.

Proof that they're not necessary:
```shell
$ zoo=v$(echo 8.3.2b1 | cut -d . -f 1,2)
$ echo $zoo
v8.3
```
Depending on your history, this may be cringe-worthy; however, the decision to not use quotes has already been made.

jputrino: Fixed. :)

ssorenso:
this is okay to merge

PR #1025, opened 2017-10-10
fix unittests in nightly

### LINKED ISSUE #1021: unit test results are not reported in TRTL.

zancas:
@pjbreaux 
Fixes #1021

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,16 +23,13 @@ pipeline {
         stage("unit"){ 
             steps {
                 sh '''
-                    # - initialize env vars
-                    . systest/scripts/init_env.sh
-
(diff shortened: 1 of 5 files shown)


zancas:
The "unit" Jenksfile stage is now _only_ run in the case of a particular Jenkins-injected parameter being set.

zancas:
TRTL does not record a build start during the "unit" stage.

zancas:
`unit_test_run_wrapper.sh` is always invoked from the "Jenkinsfile" directory.

pjbreaux:
@zancas: I though we would get this change into master, then cherry-pick the fix back. Is that not feasible right now?

PR #1035, opened 2017-10-12
update requirements files

zancas:
@jlongstaf


zancas:
@pjbreaux @jlongstaf It's critical that we note that we know of other import failures that we don't yet have unittests for, I _think_.    At any rate, these fixes cause the unittests to pass in a master environment.

zancas:
@ssorenso I'd like to discuss this test with you before we resolve this PR.

zancas:
This work will be contained in the ocata-dev upgrade.

PR #1071, opened 2017-11-02
refactor to make test_esd.py run against master

zancas:
Our new revision control strategy is to develop against master and where
necessary cherry-pick back to release branches.

This commit refactors our code so that the

f5_openstack_agent/tests/functional/esd/test_esd_migrating.py

test _runs_ (but does not pass) against master.

A primary requirement is that we update "requirements.functest.txt", in
addition to making that module correctly reference "master" in other projects
I have made it the Single Source of Truth for install_requires in setup.py,
and I have decoupled f5-openstack-agent installation from that file by
removing "-e .".

There are several other components to this refactor.

(1) I've begun moving functionality out of the top level "test" directory and
into f5_openstack_agent/tests
(2) In coordination with (1) I am changing import statements to "relative"
wherever possible.  Eventually all "in-package" import should be relative in
keeping with Python best practices:

https://www.python.org/dev/peps/pep-0328/#id9

(3) I am moving to the "filepath" library for filesystem interaction

(1) remove "-e .", this means that a separate pip install command is required
*IF* then intent is to _ALSO_ install the current project.  Of course, this
means we are allowing a development flow where the current package is _NOT_
installed for functional tests.  We should support this option.

(2) Set refspecs of remote repos to "master", this is the new normal.

(3) Add "pytest-symbols"...  it is necessary.


zancas:
@jlongstaf this passing style and unit tests

f5_openstack_agent/lbaasv2/drivers/bigip/test/test_agent_manager_LbaasAgentManager.py
```diff
@@ -228,8 +228,9 @@ def test_validate_services(self, fully_mocked_target):
         fully_mocked_target.cache.get_by_loadbalancer_id.assert_called_with(2)
         assert fully_mocked_target.cache.get_by_loadbalancer_id.call_count == 2
 
-    def itest_lbb_sync_state(self, fully_mocked_target,
-                             fully_mocked_plugin_rpc, mock_logger):
+    @pytest.mark.skip(reason="This test a swallowing a mock-derived TypeError")
```
ssorenso: Approved. I can fix this later.

f5_openstack_agent/lbaasv2/drivers/bigip/test/test_agent_manager.py
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-from f5_openstack_agent.lbaasv2.drivers.bigip import agent_manager
+from .. import agent_manager
```
ssorenso: Why are you using dots in some cases and removing them in others?

Why not just set up a blanket, consistent approach at treating all code the same?

f5_openstack_agent/tests/functional/esd/test_esd_migrating.py
```diff
@@ -18,15 +18,15 @@
 import logging
 import os
 
+from filepath import FilePath
```
ssorenso:
```python
following OpenStack convention, shouldn't it be:
import json
import logging
import os
import pytest
import requests

from copy import deepcopy
from filepath import FilePath
...
```
?
I guess that I'm surprised that style is not failing.

f5_openstack_agent/tests/functional/esd/test_esd_migrating.py
```diff
@@ -14,14 +14,15 @@
 #   limitations under the License.
 
 from copy import deepcopy
-from f5_openstack_agent.lbaasv2.drivers.bigip.icontrol_driver import \
-    iControlDriver
 import json
 import logging
 import os
+
```
ssorenso: This, too, breaks convention:
```python
import json
import logging
import os
import pytest
import requests

from filepath import FilePath

from ....lbaasv2.drivers.bigip.icontrol_driver import iConstrolDriver

from ..testlib.bigip_client import BigIpClient
from ..testlib.fake_rp import FakeRPCPlugin
from ..testlib.service_reader import LoadbalancerReader
```
This makes us closer to how OpenStack community handles imports consistently and makes F5 look better.

f5_openstack_agent/tests/functional/testlib/bigip_client.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+import re
+import urllib
+
+from f5.bigip import ManagementRoot
+from f5_openstack_agent.lbaasv2.drivers.bigip import cluster_manager
+from f5_openstack_agent.lbaasv2.drivers.bigip import resource_helper
+from f5_openstack_agent.lbaasv2.drivers.bigip import system_helper
+
```
ssorenso: Another import inconsistency.

f5_openstack_agent/tests/functional/esd/test_esd_migrating.py
zancas: I think 'filepath', 'requests', and 'pytest' need to be in a section, they are external dependencies.

There need to be _at least_ 3 sections.

One for builtins/stdlib
One for external/pip installed
One for our local code (all relative).

zancas: I'm fine with sub dividing our own imports into, e.g. app code and test code, if that seems better to you.

f5_openstack_agent/tests/functional/testlib/bigip_client.py
zancas: With respect to imports, e.g. in bigip_client, those imports should migrate to the relative import pattern.   I'll write up a complete policy for import syntax.  Since we're using "hacking" from "openstack", we won't be able to deviate too far from the community.

f5_openstack_agent/lbaasv2/drivers/bigip/test/test_agent_manager.py
zancas: All "system" packages, i.e. stdlib, and pip install-requires, e.g.:   `os`, `random`, `neutron` code shall be imported absolutely, all _within package_ code shall be imported relatively.

This sets a clear boundary between the two kinds of code:

(0) system code that is not part of this package
(1)  source code that _is_ part of this package

This enhances readability, additionally it makes it within-package imports (the ones we specify relatively, function *without respect to* whether the package has been installed).

f5_openstack_agent/tests/functional/esd/test_esd_migrating.py
jlongstaf: We should be following OpenStack style guidelines (hacking): https://docs.openstack.org/hacking/latest/user/hacking.html#imports

zancas:
@richbrowne I beg your pardon?   How about an explanation?   What is "this"?

zancas:
The only tests that have moved are the one ones formerly known as:

`https://github.com/F5Networks/f5-openstack-agent/blob/master/test/functional/neutronless/esd/test_esd.py`

All other tests remain in place.

Is there a reason we can't move those tests, which are named the same thing as another test suite?

If there is then we need to discuss the reasons.

richbrowne:
Za we cant take a change this big.

PR #1081, opened 2017-11-08
Eleven tests

zancas:
@richbrowne 

This PR adds 11 tests (2 of which are refactored old tests).

I believe this test suite shows 3 different issues, which I will file tomorrow.

To run the tests:

`cd f5-openstack-agent && pytest -v --symbols ${SYM} test/functional/neutronless/esd/test_esd.py`

### DIFF
--- a/test/functional/neutronless/esd/test_esd.py
+++ b/test/functional/neutronless/esd/test_esd.py
@@ -12,37 +12,75 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+"""These are tests of the ESD (Enhanced Services Definition) feature.
 
+ESDs are defined here:
+
+https://devcentral.f5.com/articles/customizing-openstack-lbaasv2-using-enhanced-services-definitions-25681
+
+Classes of tests in this module include:
+
+* Basic Functionality Tests:
+   These tests load and apply ESDs from an example ESD config:
+    f5-openstack-agent/etc/neutron/services/f5/esd/demo.json
+
+   Each test looks up the specific ESD appended to the "test_esd_" string in
+the test name, it then takes the resulting suffix, and prepends "f5_ESD_" to it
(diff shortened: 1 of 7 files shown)


zancas:
@ssorenso I will open issues.

Per skipping them..   I think your position makes good sense.  We need to be able to categorize types of test failure.   I think the best answer is to use `skipif` and mark the tests such that they are skipped in nightly, but run during manual testing.  Does that agree with your constraints?

etc/neutron/services/f5/esd/demo.json
```diff
@@ -1,5 +1,5 @@
 {
```
richbrowne: I would say that we should not touch this particular file since it is shipped with the product.  I realize that you are using this for test service definitions.  I would propose that you create a new file in the testdata directory used by other neutronless tests.

test/functional/neutronless/esd/test_esd.py
```diff
@@ -102,63 +104,127 @@ def esd():
...
-    validator.assert_pool_deleted(pool, None, folder)
+def test_esd_two_irules(setup):
+    infrastructure = setup
+    _apply_validate_remove_validate(infrastructure)
 
-    # delete listener
-    service = service_iter.next()
-    icontrol_driver._common_service_handler(service)
-    validator.assert_virtual_deleted(listener, folder)
+def test_esd_lbaas_ctcp(setup):
```
richbrowne: Generally add docstrings to indicate what the test is doing.

richbrowne: You may add pep 257 check on the file.

zancas: Good point.  I have added pep257 compliant docstrings.  I've also added module-specific pep257 in .travis.yml.

etc/neutron/services/f5/esd/demo.json
zancas: Acknowledged, will do.

zancas: Done.

PR #1084, opened 2017-11-10
Drive ESD tests with ESD data.

### LINKED ISSUE #1077: The "<ESD name>" _can_ collide with a neutron l7_policy name
ESD names, as described here:

https://devcentral.f5.com/articles/customizing-openstack-lbaasv2-using-enhanced-services-definitions-25681

can conflict with l7_policy names.

A simple fix is to prepend a magic string to all ESD names, e.g.:

<ESD_NAME> = "f5_ESD_"<ESD_SPECIFIC_NAME>

zancas:
This commit makes ESD tests be data driven.

It refactors test infrastructure code into:

test/functional/neutronless/esd/conftest.py

It allows developers to add a data-driven ESD
test with a single line of code, e.g.:

def test_esd_lbaas_stcp(ESD_Experiment):
    '''Stub comment'''

The "lbaas_stcp" portion informs pytest that a particular ESD is to applied,
have its application validated, be removed, have its removal be validated.

The "ESD_Experiment" specifies a "test regime" which includes service objects
to inject, and ESD configs to load.   Different regimes may specify
arbitrary Python logic to test specific behaviors.

Specify Regimes in conftest.py

ignore development file

begin refactor to extend tests

functionalize ESD_Experiment

functionalize ESD_Experiment

factor components in Experiment, and ESD_Experiment

factor infrastructure into conftest.py

move setup into conftest.py, produce pairs

first pass of ESD pairs

flake8, pep257, conftest, and test_esd_pairs

@richbrowne 
WIP #1077 
This PR doesn't fix #1077 but it does add relevant tests.


.travis.yml
```diff
@@ -24,6 +24,10 @@ script:
     - pep257 ./test/functional/neutronless/esd/test_esd.py
     - flake8 --ignore=H304 ./test/functional/neutronless/esd/test_esd.py
     - pylint -E -d E1101 test/functional/neutronless/esd/test_esd.py
```
richbrowne: I understand why you added these files to pep257, but I don't like the way it clutters up the travis file.  Could there not be a way to add this as a part of tox, and have it run as part of tox -e style?

PR #1085, opened 2017-11-10
Feature.agent ha

### LINKED ISSUE #730: agent should report as down to neutron when communication is lost to any associated BIG-IP

richbrowne:
@rbrowne @jgruber @jlongstaf @edarzins 

Fixes #730 #1088 

Where should the reviewer start?
The mock icontrol_driver needs to have connect called on it before it… …
7106199

Background Information:
What's this change do?

There was an issue in the 'create or update' workflow for virtuals on the BIG-IP because of the order of attribute processing compared to the pre-issue 730 code. The previous code looked at the Neutron provisioning_state of the Neutron listener to decide if it should try to create or update the BIG-IP virtual object. This won't work on hosts which are not in sync with provisioning. (why we changed things)

This change makes the create_listener method in listener_service make a call to the BIG-IP to see if the virtual exists and uses that to determine if a create or update is required. If the virtual service on the particular BIG-IP needs to be created.. it is created.. if it needs to be updated.. the exact same workflow that was in place before the changes is used (self.update_listener but for only the specific bigip).

The order in which LBaaS services are created in the lbaas_builder is not correct because virtuals are deployed in advance of any of its dependencies. The choosen order:

    self._assure_loadbalancer_created(service, all_subnet_hints)
    self._assure_listeners_created(service)
    self._assure_pools_created(service)
    self._assure_monitors(service)
    self._assure_members(service, all_subnet_hints)
    self._assure_pools_deleted(service)
    self._assure_l7policies_created(service)]
    self._assure_l7rules_created(service)
    self._assure_l7rules_deleted(service)
    self._assure_l7policies_deleted(service)
    self._assure_pools_configured(service)
    self._assure_listeners_deleted(service)
    self._assure_loadbalancer_deleted(service)
is problematic because the various stages will update the virtual with the elements that it knows of
without preserving the state of some of the previously run stages. The other isssue is that these build steps are not idempotent. Each time the function is run it is possible to set the virtual in a bare bones configuration before applying the policies/pools/and profiles.

This code should be reordered...

    self._assure_loadbalancer_created(service, all_subnet_hints)
    self._assure_monitors_created(service)
    self._assure_pools_created(service)
    self._assure_members(service, all_subnet_hints)
    self._assure_l7policies_created(service)]
    self._assure_listeners_created(service)

    self._assure_listeners_deleted(service)
    self._assure_l7policies_deleted(service)
    self._assure_pools_deleted(service)
    self._assure_monitors_deleted(service)
    self._assure_loadbalancer_deleted(service)


f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
```diff
@@ -375,76 +379,82 @@ def _get_lb_method(self, method):
...
-            vip.pop('vlansDisabled', None)
-        elif network_id in self.conf.common_network_ids:
-            vip['vlans'].append(
-                self.conf.common_network_ids[network_id])
-            vip['vlansEnabled'] = True
-            vip.pop('vlansDisabled', None)
+    def _add_profiles_session_persistence(self, listener, pool, vip):
+
+        protocol = listener.get('protocol', "")
+        if protocol not in ["HTTP", "HTTPS", "TCP"]:
```
jlongstaf: Need to include TERMINATED_HTTPS in the list of valid listener protocols.

f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_driver.py
```diff
@@ -15,158 +15,161 @@
...
-        """Persist backend configuratoins """
+        """Persist backend configuratoins."""
         raise NotImplementedError()
 
-    def get_stats(self, service):
-        """Get Stats for a loadbalancer Service """
-        raise NotImplementedError()
+    def generate_capacity_score(self, capacity_policy):
+        """Generate the capacity score of connected devices."""
+        raise NotImplemented
```
jlongstaf: Shouldn't we raise NotImplementedError() and not NotImplemented? NotImplemented is a Python constant (AFAIK), and the exception class name is NotImplementedError. Someone correct me if I'm wrong. I know this is from the previous code and not introduced in these changes, but we might want to correct this.

f5_openstack_agent/lbaasv2/drivers/bigip/pool_service.py
```diff
@@ -53,8 +56,28 @@ def create_pool(self, service, bigips):
         :param bigips: Array of BigIP class instances to create pool.
         """
         pool = self.service_adapter.get_pool(service)
+        error = None
+
         for bigip in bigips:
-            self.pool_helper.create(bigip, pool)
+            try:
```
edarzins: Would be nice if you could abstract this pattern of "perform operation, catch exceptions, log error, return result" -- which you did in CCCL. Or maybe that's just something we defer until CCCL integration?

f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
```diff
@@ -134,186 +158,120 @@ def _assure_listeners_created(self, service):
...
+
+                if error:
                     loadbalancer['provisioning_status'] = \
                         plugin_const.ERROR
                     listener['provisioning_status'] = plugin_const.ERROR
-                    raise f5_ex.VirtualServerCreationException(err.message)
-            self._set_status_as_active(listener)
+                else:
+                    listener['provisioning_status'] = plugin_const.ACTIVE
+                    listener['operating_status'] = lb_const.ONLINE
```
jlongstaf: Shouldn't he operating_status depends on the value of admin_state_up? The VS could have been created with admin_state_up = False.

jgruber:
There is no default 'device_interfaces' attribute on bigip objects. This creates a error:

`2017-11-15 12:09:22.293 29901 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.icontrol_driver [-] Invalid agent configuration: '' object has no attribute 'device_interfaces'
2017-11-15 12:09:22.293 29901 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.utils [-] get_all_deployed_loadbalancers request b47f814a-aa38-45df-9920-ebfba89e065c FAILED
2017-11-15 12:09:22.294 29901 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager [-] Unable to sync state: '' object has no attribute 'device_interfaces'
`

jgruber:
The new workflow obviously precludes the use of ESDs. The _assure_l7policy_created only handles non-esd policies with rules attached.

jgruber:
In my testing I can create situations where SNAT translation address, SNAT pools are not deleted, and thus the route domain delete and tenant folder delete don't complete. This creates a problem with declarative orchestrations where the entire DELETE process, which includes the Neutron network and subnets, fails with stranded Neutron ports for SNATs, SelfIPs making the Neutron subnet delete impossible.

f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_service.py
```diff
@@ -15,159 +15,94 @@
...
+                    error = f5_ex.L7PolicyDeleteException(err.message)
+            except Exception as err:
+                LOG.exception(err)
+                error = f5_ex.L7PolicyDeleteException(err.message)
+
+            if error:
+                LOG.error("L7 Policy deletion error: %s",
+                          error.message)
+
+            return error
```
jlongstaf: This return statement is inside the for loop. Should it be outside the loop?

f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
```diff
@@ -134,186 +158,120 @@ def _assure_listeners_created(self, service):
...
-
-    def _get_pool_listeners(self, service, pool_id):
-        pools_listeners = []
-        for listener in service['listeners']:
-            if listener['default_pool_id'] == pool_id:
-                pools_listeners.append(listener)
-        return pools_listeners
+                else:
+                    pool['provisioning_status'] = plugin_const.ACTIVE
+                    pool['operating_status'] = lb_const.ONLINE
```
jlongstaf: Same as for listener: pool can be created with admin_state_up = False, and then should have an operating_status of OFFLINE.

f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
```diff
@@ -49,36 +51,54 @@ def __init__(self, conf, driver, l2_service=None):
             self.service_adapter
         )
         self.l7service = l7policy_service.L7PolicyService(conf)
+        self.clean_wrapper_policy = True
```
jlongstaf: I see how clean_wrapper_policy is used at line 504, but how is this value set/unset?

richbrowne: It's not.  I need to consider how this should be done.

f5_openstack_agent/lbaasv2/drivers/bigip/pool_service.py
richbrowne: Thanks Erik.  My experience with this lead me to implementing CCCL the way it is, but it's a stretch to redo that here.

f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_driver.py
richbrowne: This was on there before.  I don't recall changing this.

PR #1094, opened 2017-11-21
Fixes tracker to use dynamic, symbols settings

ssorenso:
Issues:

Problem:
* Symbols should be used, if available

Analysis:
* This sets a static check for the symbols
  * If there are, and it has the right symbol for mgmt ip...
    * Then it uses this
  * else, the static config is attempted

Tests:
This is a test suite conftest fixture change

@richbrowne 
Static ssh config file linking if a symbols file is present/available

#### What's this change do?
This change will check if a pytest.symbols is set, and if so, use the bigip settings under that.  If not, then it will attempt to use the static ssh config file on bastion

bigip_interaction.py

#### Any background context?
It was assuming before that the neutron test would be run from within bastion.  This will now run anywhere where there is access to a bigip as long as the symbols file is specified for pytest.

### DIFF
--- a/test/functional/neutronless/bigip_interaction.py
+++ b/test/functional/neutronless/bigip_interaction.py
@@ -74,7 +74,7 @@ list sys folder recursive one-line" | cut -d " " -f3 |
 while read f; do echo \"====================\";
 echo \"Folder $f\"; tmsh -c "cd /$f; list\"; done;
 exit
-EOF'''.format(ssh_cmd)
+EOF'''
     __ucs_cmd_fmt = "{} tmsh {} /sys ucs /tmp/backup.ucs"
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


ssorenso:
Test Results:
```shell
$ py.test --symbols /git/clones/testenv-openstack-bigip/testenv_symbols/testenv_symbols/testenv_symbols.json ./loadbalancer/
======================================= test session starts =======================================
platform darwin -- Python 2.7.10, pytest-3.2.5, py-1.5.2, pluggy-0.4.0
rootdir: /git/clones, inifile:
plugins: symbols-0.1.1, cov-2.5.1, f5-sdk-2.3.3
collected 13 items

loadbalancer/test_f5_common_networks.py Es
loadbalancer/test_loadbalancer.py ..
loadbalancer/test_loadbalancer_admin_state.py .
loadbalancer/test_loadbalancer_disconnected.py ..
loadbalancer/test_loadbalancer_opflex.py ..
loadbalancer/test_loadbalancer_snat.py ...
loadbalancer/test_snat_common_network.py .
```

PR #1101, opened 2017-11-28
refactor neutronless test infrastructure

zancas:
@janakimeyyappan 
Issues:
Fixes:  This is a test issue.

Problem: In the previous iteration of the neutronless test architecture all
configuration was accomplished with a tall stack of fixtures.  This worked for
a given set of tests, but proved to be very difficult to modify.

Analysis:  In order to make it easier to write different kinds of tests
the bulk of configuration has been migrated into a standard Python class.  This
class is instantiated within a fixture, guaranteeing that entry into the test
system still conforms to the behaviors enforced by pytest.

The result is a much more flexible system that maintains the feature set of
pytest.

Tests:  I ran the tests inside neutronless/esd

### DIFF
--- a/test/functional/neutronless/conftest.py
+++ b/test/functional/neutronless/conftest.py
@@ -12,28 +12,189 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+"""In addition to pytest fixtures this module also contains general classes.
 
-import json
-import os
-import pytest
-import re
-import traceback
-import sys
+    These classes are TestConfig, and Resource4TestTracker.
 
+    TestConfig:  This is a Python class designed to contain the configurable
+state and methods generally needed during testing.  The general approach is
+to provide a configuration specific to some test needs via the TestConfig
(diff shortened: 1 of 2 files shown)


test/functional/neutronless/conftest.py
```diff
@@ -12,29 +12,190 @@
...
+to provide a configuration specific to some test needs via the TestConfig
+objects methods (primarily, __init__, but also others e.g. "load_esd").
+
+   The expected caller of the TestConfig object is an Experiment fixture, where
+the experiment instantiates, and configures, a particular TestConfig instance, 
+before handing off to the test code.   Examples of "Experiments" can be found
+in esd/conftest.py.
+"""
+
+import collections
```
ssorenso: Community-driven requirement would be:
```python
import collections
import json
import os
import pytest
import re
import sys
import traceback

from collections import deque
from collections import namedtuple
from copy import deepcopy
from inspect import currentframe as cf
from inspect import getframeinfo as gfi
from inspect import getouterframes as gof
```

Technically, though, the hacking guidelines would have you import at the module level and not at the function/method level... this is longstanding code, though... so I'm not sure if it's appropriate to fix it here... or not.

Issue #1117, opened 2017-12-01
Service validation error after upgrading from LIberty to Newton release


jlongstaf:
#### Agent Version
10.1.0

#### Operating System
Unknown.

#### OpenStack Release
Newton

#### Description
The following error is reported in the agent log. This error happens repeatedly:
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager [-] Service validation error: object of type 'NoneType' has no len()
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager Traceback (most recent call last):
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager   File "/usr/lib/python2.7/site-packages/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py", line 575, in validate_service
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager     if not self.lbdriver.service_exists(service) or \
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager   File "/usr/lib/python2.7/site-packages/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py", line 772, in service_exists
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager     return self._service_exists(service)
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager   File "/usr/lib/python2.7/site-packages/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py", line 1340, in _service_exists
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager     monitor = self.service_adapter.get_healthmonitor(svc)
2017-11-22 03:20:56.788 87792 ERROR f5_openstack_agent.lbaasv2.drivers.bigip.agent_manager   File "/usr/lib/python2.7/site-packages/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py", line 169, in get_healthmonitor
[...]

PR #1122, opened 2017-12-04
Fix service validation error after upgrading from LIberty to Newton release

jlongstaf:
@richbrowne 
#1117 Service validation error after upgrading from LIberty to Newton release 

#### What's this change do?
Instead of checking for valid name and description strings using len(), substituted using more general truth value testing (.e.g, if value instead of if len(value)). In this way, values with None type and values with empty string are both handled without exception.

service_adapter.py

#### Any background context?
Normally we don't expect None types for resource values. But this can happen when a new database column is initialized to NULL, as is the case when upgrading OpenStack releases (health monitor names not defined in Liberty and added in Mitaka release).


richbrowne:
Go ahead and create PR's for mitaka and stable/newton.

richbrowne:
Accidentally, closed PR.  Reopened.

PR #1125, opened 2017-12-04
Fixing bigip_interaction.py to appropriately clean

ssorenso:
Problem:
* Fixes test bug where bigip_interaction.py did not clean up after tests
  * This is largely due to inappropriately handled variables and content

Solution:
* Grabs the content from the BIG-IP once post-test
  * This is then stored appropriately under the dirty-config file
  * This file is then appropriately diff'ed with the "clean" one
  * The resulting diff cmd call results in a exit status of (1)
  * Handling this by then handling the resulting RuntimeError
    * Transforming this into the appropriate AssertionError

Tests:
Have perform the following tests:
* file ./test/functional/neutronless/loadbalancer/test_purge_folder.py
  * Comment line `sh.purge_folder_contents(bigip.bigip, folder)`
  * Comment line `sh.purge_folder(bigip.bigip, folder)`
* Run py.test against this file
  * Expected to fail at this point
* Check BIG-IP config to assure that state is clean
* Revert changes to the test_purge_folder.py and re-test
  * Expect a passing result

@richbrowne @jlongstaf @janakimeyyappan 
BIG-IP cleanup issues via the tracker in the agent's functional tests.

#### What's this change do?
* conftest.py missing the importation and use of the bigip_interaction.py for its fixture
* Handle the dirty contents properly from the BIG-IP (fixes a bug here)
  * Was always coming back as 'clean' unintentionally
* Handles the diff command call appropriately
  * Catches the RuntimeError from calling _check_results() method
  * Handles then returning these results properly and cleaning up the BIG-IP by restoring from the backup ucs file

conftest.py, which was commented out over to bigip_interaction.py changes

#### Any background context?
@janakimeyyappan ran into a scenario in her purge testing that the BIG-IP was not getting auto-cleaned up.  These changes seem like they were fixes I already implemented earlier...  though I can't now find a record of it in one of my old branches.

### DIFF
--- a/test/functional/neutronless/bigip_interaction.py
+++ b/test/functional/neutronless/bigip_interaction.py
@@ -139,9 +139,7 @@ EOF'''
             * Generate a diff file against the polluted config
         """
         try:
-            with open(cls.config_file.format(my_epoch)) as fh:
-                content = fh.read()
-            diff_file = cls.__collect_diff(content, test_method)
+            diff_file = cls.__collect_diff(test_method)
             os.remove(diff_file)
         except AssertionError as err:
             cls.__restore_from_backup()
@@ -165,7 +163,7 @@ EOF'''
         return diff_file
 
     @classmethod
-    def __collect_diff(cls, clean, test_method):
+    def __collect_diff(cls, test_method):
         """An internal method"""
         dirty_file = cls.dirty_file.format(test_method, my_epoch)
         dirty_content = cls._get_current_bigip_cfg()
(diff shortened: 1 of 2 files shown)


ssorenso:
We may want to just transfer these files over manually.

test/functional/neutronless/bigip_interaction.py
```diff
@@ -159,17 +165,19 @@ def _collect_diff(cls):
         return diff_file
 
     @classmethod
-    def __collect_diff(cls, result, test_method):
+    def __collect_diff(cls, clean, test_method):
```
richbrowne: I don't see where the 'clean' parameter is used.

richbrowne:
This is great.

Issue #582, opened 2017-02-14
Deleting pool with members leaves member network config on bigip


richbrowne:
When a user calls lbaas-pool-delete with members, any member networking deployed when the member was created is left on the bigip.  This causes resource leaks.

PR #1142, opened 2017-12-06
Fix deleting pool with members leaves member network config on bigip

jlongstaf:
@richbrowne 
#582 Deleting pool with members leaves member network config on bigip

#### What's this change do?
Rearranges how pool members are deleted and when subnet hints are processed,
to ensure that subnets and associated network config is cleaned up after deleting
a pool that has members.

lbaas_builder.py

#### Any background context?
No.


richbrowne:
Please create the PR's for mitaka and newton and I will merge all at the same time.

PR #1144, opened 2017-12-06
Master.bigip tracker fixes

ssorenso:
@richbrowne @jlongstaf @larkinkevin 
* BIG-IP config OoO resulting in improper triggering on non-crufted config
* Testers workflow degraded due to constant tracking
  * Tester can add a 'no_bigip_tracker' value in the symbols file to turn off tracking
* Testers can transparently use symbols or testenv for testing
  * This will handle properly setting up the ssh command used by the tracker
* Testers can transparently ignore ssh prompting
  * **It is still up to the tester to share a ssh-added key for their own custom BIG-IP deployments**

#### What's this change do?
* Adding additional diff logic checking to assure that the file size of pre-test config matches post-test config
  * This is not perfect and may allow appropriately-triggered diffs to be ignored!
  * **NOTE:** Further tracking will be added to this later and this work is in the planning stages
* Added 'no_bigip_tracker' flag checking for symbols that bypasses tracking
* Refactored ssh_cmd variable setting in the `bigip_interactions.py` library
* Fixed testenv symbols referencing
* Fixed pytest symbols referencing
* Dumping known_host crufting to `/dev/null`
* Ignoring known_host key checking

`bigip_interactions.py` in `neutronless`, agent func-tests

#### Any background context?
All the problems with tracking...

### DIFF
--- a/test/functional/neutronless/bigip_interaction.py
+++ b/test/functional/neutronless/bigip_interaction.py
@@ -185,7 +185,8 @@ EOF'''
 
         test_name := the name of the test currently in tearDown
         """
-        cls._resulting_bigip_cfg(test_name)
+        if not hasattr(pytest.symbols, 'no_bigip_tracking'):
+            cls._resulting_bigip_cfg(test_name)
 
     @classmethod
     def backup_bigip_cfg(cls):
@@ -194,7 +195,9 @@ EOF'''
         This method will store a backup of the BIG-IP's configuration on the
         BIG-IP for later restoration.
         """
-        if not os.path.isfile(cls.config_file.format(my_epoch)):
+        if hasattr(pytest.symbols, 'no_bigip_tracking'):
+            pass
+        elif not os.path.isfile(cls.config_file.format(my_epoch)):
             cls.__exec_shell(
                 cls.__ucs_cmd_fmt.format(cls.ssh_cmd, 'save'), True)
             cls._get_existing_bigip_cfg()
--- a/test/functional/neutronless/bigip_interaction.py
+++ b/test/functional/neutronless/bigip_interaction.py
@@ -22,12 +22,7 @@ import subprocess
 from collections import namedtuple
 from time import sleep
 
-ssh_cfg = os.environ.get('ssh_cfg', None)
-if not ssh_cfg:
-    ssh_cfg = '/home/ubuntu/testenv_symbols/testenv_ssh_config'
 my_epoch = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
-current_test = ''  # should contain the current tests's name
-
 
 """Allows test interaction with the BIG-IP
 
@@ -62,12 +57,12 @@ class BigIpInteraction(object):
     This class assumes that you are in testenv and that there is a BIG-IP
     configured for test runner's use.
     """
-    config_file = "/tmp/agent_only_bigip_{}.cfg"
+    __current_test = ''
     diff_file = '/tmp/agent_only_bigip_{}_{}.diff'
     dirty_file = '/tmp/agent_only_bigip_dirty_{}_{}.cfg'
+    config_file = '/tmp/agent_only_bigip_{}.cfg'
     _lbs_to_delete = []
-    ssh_cmd = \
-        str("ssh -F {} openstack_bigip").format(ssh_cfg)
+    ssh_cmd = ''
     __extract_cmd = '''{} << EOF
 tmsh -c \"cd /;
 list sys folder recursive one-line" | cut -d " " -f3 |
@@ -180,13 +175,6 @@ EOF'''
         return diff_file
 
     @classmethod
-    def check_symbols(cls):
-        if hasattr(pytest, 'symbols') and \
-                hasattr(pytest.symbols, 'bigip_mgmt_ip_public'):
-            cls.ssh_cmd = "ssh {}@{}".format(
-                pytest.symbols.bigip_ssh_username,
-                pytest.symbols.bigip_mgmt_ip_public)
-
     @classmethod
     def check_resulting_cfg(cls, test_name=current_test):
         """Check the current BIG-IP cfg agianst previous Reset upon Error
@@ -197,7 +185,6 @@ EOF'''
 
         test_name := the name of the test currently in tearDown
         """
-        cls.check_symbols()
         cls._resulting_bigip_cfg(test_name)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


test/functional/neutronless/bigip_interaction.py
```diff
@@ -22,11 +22,7 @@
 from collections import namedtuple
 from time import sleep
 
-ssh_cfg = os.environ.get('ssh_cfg', None)
-if not ssh_cfg:
-    ssh_cfg = '/home/ubuntu/testenv_symbols/testenv_ssh_config'
 my_epoch = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
-current_test = ''  # should contain the current tests's name
```
ssorenso: This has to be modified to handle the missing variable here.

ssorenso: This is handled now properly.

ssorenso:
@larkinkevin @jlongstaf @richbrowne 
This is ready for review.

test/functional/neutronless/bigip_interaction.py
```diff
@@ -165,7 +161,8 @@ def _collect_diff(cls):
     @classmethod
     def __collect_diff(cls, test_method):
         """An internal method"""
-        dirty_file = cls.dirty_file.format(test_method, my_epoch)
+        dirty_file = cls.dirty_file.format(my_epoch, test_method)
+        sesson_file = cls.config_file.format(my_epoch)
```
larkinkevin: is this supposed to be "session_file"?

ssorenso: Why... yes, yes it is...

test/functional/neutronless/bigip_interaction.py
```diff
@@ -176,19 +173,13 @@ def __collect_diff(cls, test_method):
         try:
             cls.__check_results(result)
         except RuntimeError:
-            raise AssertionError(diff_file)
+            if os.path.getsize(diff_file) != os.path.getsize(sesson_file):
```
larkinkevin: Shouldn't this test (of the file sizes) be part of "__check_results"?

test/functional/neutronless/bigip_interaction.py
```diff
@@ -197,26 +188,74 @@ def check_resulting_cfg(cls, test_name=current_test):
 
         test_name := the name of the test currently in tearDown
         """
-        cls.check_symbols()
-        cls._resulting_bigip_cfg(test_name)
+        if test_name:
+            cls.__current_test = test_name
+        else:
+            test_name = cls.__current_test
+        if not hasattr(pytest.symbols, 'no_bigip_tracking') and cls.ssh_cmd:
```
larkinkevin: This if-statement would be easier to read if the outer if tested for hasattr and the inner if tested for cls.ssh_cmd.

test/functional/neutronless/bigip_interaction.py
```diff
@@ -197,26 +188,74 @@ def check_resulting_cfg(cls, test_name=current_test):
...
             cls.__exec_shell(
                 cls.__ucs_cmd_fmt.format(cls.ssh_cmd, 'save'), True)
             cls._get_existing_bigip_cfg()
+        elif not hasattr(pytest.symbols, 'no_bigip_tracking') and \
+                not cls.ssh_cmd:
+            raise EnvironmentError("Broken state... cannot track without "
+                                   "ssh_cmd!")
+
 
+def BEGIN():
```
larkinkevin: This doesn't look like a valid function name.  You sure it will get past the linter/style-checks?

test/functional/neutronless/bigip_interaction.py
```diff
@@ -197,26 +188,74 @@ def check_resulting_cfg(cls, test_name=current_test):
...
+    environment variables.
+
+    User Notes: if it is found that bigip config tracking is unnecessary and is
+    reducing or impacting production time, an added 'no_bigip_tracker' variable
+    can be added to the --symbols <file>.  In this instance, this library will
+    not execute, and all fixtures that use this library will not perform
+    bigip config tracking.
+
+    WARNING: Nightly will perform this tracking; thus, a user does no justice
+    if linging config is later discovered.  It is; therefore, still recommended
```
larkinkevin: linging --> lingering?
"does no justice" --> "does themselves no justice"?

test/functional/neutronless/bigip_interaction.py
ssorenso: no.  The function `__check_results` is used as a means of checking all shell command's responses, not just for the diff shell command execution.  I would rather this guy remain as simple as he is.

test/functional/neutronless/bigip_interaction.py
```diff
@@ -197,26 +188,74 @@ def check_resulting_cfg(cls, test_name=current_test):
...
+    to run this as a last step to assure proper config tracking is assured.
+    """
+    ssh_options = """ssh -o StrictHostKeyChecking=no \
+    -o UserKnownHostsFile=/dev/null \
+    {}"""
+    ssh_host_specific_fmt = "{}@{}"
+    ssh_cmd = ''
+    hostname = ''
+    username = ''
+    if 'TESTENV_AGENT_BRANCH' in os.environ:
```
larkinkevin: I think we should use symbols as the sole input interface for the openstack tests.  That is/was the primary purpose of the pytest-symbols plugin.

The only time testenv should come up in conversation is when we talk about whether or not it was used to produce the symbols file.

The tests themselves shouldn't have to know about (or mention) testenv.

test/functional/neutronless/bigip_interaction.py
```diff
@@ -168,7 +168,7 @@ def __collect_diff(cls, test_method):
             fh.write(dirty_content)
         diff_file = cls.diff_file.format(test_method, my_epoch)
         cmd = "diff -u {} {} > {}".format(
-            cls.config_file.format(my_epoch), dirty_file, diff_file)
+            sesson_file, dirty_file, diff_file)
```
larkinkevin: session_file?

larkinkevin:
FYI, it's not immediately obvious to the reader that "OoO" means "out of order".

test/functional/neutronless/bigip_interaction.py
```diff
@@ -197,26 +188,74 @@ def check_resulting_cfg(cls, test_name=current_test):
...
-            cls.check_symbols()
+        cls.__current_test = test_name
+        if not os.path.isfile(cls.config_file.format(my_epoch)) and \
+                not hasattr(pytest.symbols, 'no_bigip_tracking') and \
+                cls.ssh_cmd:
             cls.__exec_shell(
                 cls.__ucs_cmd_fmt.format(cls.ssh_cmd, 'save'), True)
             cls._get_existing_bigip_cfg()
+        elif not hasattr(pytest.symbols, 'no_bigip_tracking') and \
+                not cls.ssh_cmd:
```
larkinkevin: Is this "elif" necessary, given the contents of the "BEGIN" function (ie. it seems like BEGIN is intended to prevent this condition)?

test/functional/neutronless/bigip_interaction.py
```diff
@@ -197,26 +188,74 @@ def check_resulting_cfg(cls, test_name=current_test):
...
 
         This method will store a backup of the BIG-IP's configuration on the
         BIG-IP for later restoration.
         """
-        if not os.path.isfile(cls.config_file.format(my_epoch)):
-            cls.check_symbols()
+        cls.__current_test = test_name
+        if not os.path.isfile(cls.config_file.format(my_epoch)) and \
+                not hasattr(pytest.symbols, 'no_bigip_tracking') and \
+                cls.ssh_cmd:
```
larkinkevin: Can this "if" be simplified to something lke:
`if not hasattr(pytest.symbols): return
cls.__exec_shell(...`

ssorenso:
Fixed @larkinkevin's requested changes

test/functional/symbols.json.example
```diff
@@ -1,6 +1,7 @@
 {
     "bigip_mgmt_ip": "10.190.7.183",
     "bigip_mgmt_ip_public": "10.190.7.183",
+    "bigip_ssh_username": "root",
```
larkinkevin: @richbrowne if you don't want the default symbols file to be updated to reflect this new required input, then you'll have to change the default behavior of the bigip-tracker feature (to make it "opt in", so it's not "on" by default).

With the current setup, if the bigip-tracker feature is enabled, then the bigip_ssh_username symbol is required.

test/functional/neutronless/bigip_interaction.py
```diff
@@ -231,31 +223,26 @@ def BEGIN():
...
     else:
-        raise EnvironmentError("Cannot perform tests without symbols or "
-                               "expected testenv environment!")
+        raise EnvironmentError("Cannot perform tests without symbols!")
     ssh_cmd = ssh_options.format(
         ssh_host_specific_fmt.format(username, hostname))
     BigIpInteraction.ssh_cmd = ssh_cmd
 
 
-BEGIN()
```
larkinkevin: In the commit message:
"chanes" --> "changes"?

or just "Apply code review comments."?

larkinkevin:
Looks good to me.

Did the functional tests run okay with these changes?

ssorenso:
Yes, the functional tests ran just fine in testenv with these changes.

test/functional/neutronless/bigip_interaction.py
```diff
@@ -22,12 +22,8 @@
 from collections import namedtuple
 from time import sleep
 
-ssh_cfg = os.environ.get('ssh_cfg', None)
-if not ssh_cfg:
-    ssh_cfg = '/home/ubuntu/testenv_symbols/testenv_ssh_config'
 my_epoch = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
-current_test = ''  # should contain the current tests's name
-
+do_tracking = True
```
richbrowne: This is not needed is it?

PR #1173, opened 2017-12-15
Fix test failures for test_network.py with BIG-IP version 11.5.4

jlongstaf:
#### What's this change do?
Changes how filters are constructed in the test. The existing filter strings work only for versions greater than 11.6, which causes the tests to fail against 11.5.4. This change uses get_filter() defined in utils.py, same as what the code does, which creates correct filters for different versions of BGI-IP. Additional changes are just formatting mods to pass flake8.

Start with line 93 and look at how the filter is constructed.

#### Any background context?
All tests except the ARP tests will now pass in 11.5.4.


janakimeyyappan:
This needs to be cherry picked to other branches ? mitaka and master ?

PR #1183, opened 2018-01-09
Fixes delete_l7policy for clusters

### LINKED ISSUE #1182: [BUG] l7policy is not getting deleted in clustered BIG-IP environment
* Title: Found in Test l7policy is not getting deleted in clustered BIG-IP environment
* Attachments: 
* Details: [f5_openstack_agent.lbaasv2.drivers.bigip.l7_service.py](https://github.com/F5Networks/f5-openstack-agent/blob/master/f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_service.py#L93)'s delete_l7policy method is returning after either slave or master BIG-IP deletes the policy (pre-maturely).

#### OpenStack Release
All supported (except Liberty)

#### Description
Clustered BIG-IP environment in OpenStack where l7policies are implemented will leave the l7policy after a delete on either the master or the slave BIG-IP.

##### Required Conditions

1. The environment has 2 BIG-IP's (or more) in a cluster
2. Neutron has been assigned to create a l7policy on a listener
3. Neutron has been assigned to create one or more l7rules on the l7policy
4. Neutron has been told to delete the l7rules and policy

##### Impact

- When neutron is assigned to delete the listener and loadbalancer (and all subsequent objects under the loadbalancer that might exist), the loadbalancer is set into an ERROR provisioning status
- The loadbalancer object can then still be deleted from a neutron level; however, it will remain on either the master or slave BIG-IP

##### Workaround
**When the policy has been deleted, but not the loadbalancer**
After the deletion of the l7policy on the master or slave BIG-IP log into the one that still has the l7policy via the UI.  Go then to `Local Traffic` > `Virtual Servers` and select the Project_ partition that corresponds to the subnet's `tenant_id` at the top-right corner of the BIG-IP's browser UI.  Then select `Policies` under `Local Traffic` > `Virtual Servers`, and the user should see a policy listed named `wrapper_policy_<subnet's tenant_id>`.  Verify that this setting exists on one BIG-IP and not the other by logging into the other BIG-IP in the cluster and navigating to this same section.  The user should not see the `wrapper_policy_<subnet's tenant_id>` on this mirroring page!  Select this policy's check-box and click `Delete` on the BIG-IP that has this policy.

**When the loadbalancer has also been deleted**
Follow the section **When the policy has been deleted, but not the loadbalancer** up to the verification.  Here, you should not see the partition `Project_<subnet's tenant_id>` as an option under partitions on the split-brained BIG IP.  This is 'okay' and expected by this point.  Go ahead and delete the `wrapper_policy_<subnet's tenant_id>` as the instructions above state.  Then, additionally, go under the `System` setting and then `Users`.  Under the `Partition List` tab, select the `Project_<subnet's tenant_id>` option's check-box and click `Delete`.

#### Agent Version
All latest Mitaka releases after 9.3.3 and Newton releases after 10.0.1b.  Does not impact 9.3.3 or 10.0.1b

#### Fix
Review the release notes and install the agent version that has this ticket fixed.  If the user has already run into this bug's state, it is required to perform the actions under Workaround to fix this issue.

#### Operating System
This issue is OS independent and impacts all installations under these circumstances.

#### OpenStack Release
Mitaka Newton

#### Deployment
2 or more BIG IP's in a cluster in an OpenStack environment where l7policy has been deployed on a listener via neutron.

ssorenso:
Issues:
Fixes: #1182

Problem:
* Clustered, 2 or more BIG IP's where l7policy has been deployed
  * Only if deployed via neutron in OpenStack
  * The LocalTraffic > Virtual Server > Policies
    * wrapper_policy_<subnet's tenant_id> policy will remain on...
      * slave or...
      * master

Analysis:
* This moves the `return error` line to outside of the BIG IP loop
  * This means that the l7policy delete change will occur on both master
    and slave

Tests:
Discovered by testing multi-BIG IP clusters in tempest tests

@richbrowne 
Fixes #1182 

#### What's this change do?
Moves the `return error` outside of the loop of BIG IP's.

`l7policy_service.py`

#### Any background context?
This was discovered in nightly testing with multiple BIG IP's in a cluster.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_service.py
@@ -90,7 +90,7 @@ class L7PolicyService(object):
                 LOG.error("L7 Policy deletion error: %s",
                           error.message)
 
-            return error
+        return error
 
     def build_policy(self, l7policy, lbaas_service):
         # build data structure for service adapter input


f5_openstack_agent/lbaasv2/drivers/bigip/l7policy_service.py
```diff
@@ -90,7 +90,7 @@ def delete_l7policy(self, f5_l7policy, bigips):
                 LOG.error("L7 Policy deletion error: %s",
                           error.message)
 
-            return error
+        return error
```
amudukutore: Is there a python linter that could have caught this problem (i.e., function did not call return in all cases)?

richbrowne: Py.lint might have caught the error.

PR #1200, opened 2018-01-13
Master.orphaned healthmonitor existential crisis

ssorenso:
@richbrowne @jlongstaf 
Orphaned healthmonitors sticking around...

#### What's this change do?
* Fixes some improper l7policy purging attribute references
* Adds more verbose logging to purge_orphaned_health_monitors
* Handles healthmonitor id/name referencing between
* Fixes problem of orphaned healthmonitors not being purged properly
  * And their pre-mature attempts when they are not orphaned...

`icontrol_driver.iControlDriver.get_all_deployed_pools()`

#### Any background context?
Before, we would see the periodic task fail in `icontrol_driver.iControlDriver.get_all_deployed_listeners()` due to improper attribute references for l7policy extraction (when an l7policy does not exist on the listener). 

Then there was an issue of the healthmonitor deletion attempt when it was still being referenced by a pool (when it should be there).  Then properly being deleted (when it should not be there).

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
@@ -606,7 +606,7 @@ class LbaasAgentManager(periodic_task.PeriodicTasks):  # b --> B
                       % list(known_services))
 
         except Exception as e:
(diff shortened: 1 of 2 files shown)


richbrowne:
Please create versions for Mitaka and Newton.

PR #1222, opened 2018-01-18
Fixing purge orphaned l7policy feature

### LINKED ISSUE #1209: Orphaned L7policy objects are not getting purged when no reference in Neutron DB
* Title: Orphaned L7policy objects are not getting purged when no reference in Neutron DB
* Attachments: None (yet)
* Details: Currently, the `agent_manager.purge_orphaned_l7_policies` method is attempting to delete a `policy_<li_id>` object, when it should be `wrapper_policy_<li_id>`.  The second part of this issue is that it's not matching what policies are referencing the listeners and what are not.

##### Suggested bug type:
BUG - this is a product/feature bug

#### OpenStack Release
mitaka/newton pre-release branches

#### Description
purge_orphaned_l7_policies is failing due to...
* Attempting to purge the l7policy too soon
  * `icontrol_driver.iControlDriver.get_all_deployed_listeners` method is not returning the proper l7policy references off of the BIG-IP with the listeners they reference
  * The naming change in `icontrol_driver.iControlDriver.get_all_deployed_l7policys` where the `wrapper_` portion of `wrapper_policy_<li_id>` is getting removed to the sub-string `policy_<li_id>` 
    * Due to the `Project_` prefix sub-string length being removed unintentionally

#### Agent Version
Pre-release Newton/Mitaka - this does not impact any official release

#### Operating System
OS-independent

#### OpenStack Release
Newton/Mitaka

#### Deployment
These problems should be independent of this; however, it was discovered in a 2 BIG-IP cluster environment with a single network controller.

ssorenso:
Issues:
Fixes #1209

Problem:
* Orphaned l7policies were not getting purged
  * Conditions where this would occur:
    * l7policy added to listener on neutron config
    * The BIG-IP is taken down (temporarily)
    * The l7policy is removed in then neutron config
    * The BIG-IP is brought back online
  * Symptoms
    * The policies on the virtual remain on the BIG-IP indefinitely

Analysis:
* This will query the Neutron config via the f5lbaasdriver library
  * Via RPC
  * If neutron no longer has any l7policies referencing the listener...
    * Then the policy pointing to the virtual are de-referenced
    * The policy is deleted on the BIG-IP
* This will grab a list of policies that are referenced by virtual
  * If there is a policy that is no longer referenced by a virtual...
    * Then the policy is deleted on the BIG-IP
  * This covers the scenario where the orphaned listener is purged

Tests:
Look to systest test

@richbrowne 
Fixes #1209 

#### What's this change do?
* If the BIG-IP is unaware of the fact that...
  * On Neutron, no l7policies exist anymore on a listener...
* Then the agent will execute a purge by...
  * Unlinking the virtual (listener) on the BIG-IP from the policy (l7policy)
  * Deleting the policy (l7policy) off of the BIG-IP

#### What's this change not do?
* If more than 1 l7policies exist on a listener
  * And if one of these l7policies (and its rules) are deleted off of the network controller
  * And a BIG-IP in the cluster is down or made to be unaware of this change
* Then the agent cannot know to delete this l7policy and consolidate it on the BIG-IP policy that is linked to the listener in question

##### Workaround
1. Make sure that the BIG-IP in question remains online...
2. Change the name of the listener temporarily to include an arbitrary sub-string (or remove an arbitrary substring from it) on the Controller
3. Return that substring on the controller

###### How this works
What this does is cause the neutron_lbaas to prompt a listener update in the form of a service object on the f5-openstack-agent.  This will include all present l7policies and re-sync the whole state on the BIG-IP.  Due to the current design of the f5-openstack-agent, there is no means to handle this specific case, **and it is recommended**, though not required to keep only 1 l7policy per listener, if at all possible.  If not, please be aware of this work around.

agent_manager's `purge_orphaned_l7_policys()` method

#### Any background context?
Please see what it does do versus what it doesn't...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
@@ -881,21 +881,41 @@ class LbaasAgentManager(periodic_task.PeriodicTasks):  # b --> B
     def purge_orphaned_l7_policys(self, policies):
         """Deletes hanging l7_policies from the deleted listeners"""
         policies_used = set()
-        listeners = self.lbdriver.get_all_deployed_listeners()
+        listeners = self.lbdriver.get_all_deployed_listeners(
+            expand_subcollections=True)
         for li_id in listeners:
             policy = listeners[li_id]['l7_policy']
             if policy:
                 policy = policy.split('/')[2]
             policies_used.add(policy)
+        has_l7policies = \
+            self.plugin_rpc.validate_l7policys_state_by_listener(
+                listeners.keys())
         # Ask Neutron for the status of all deployed l7_policys
         for policy_key in policies:
             policy = policies.get(policy_key)
+            purged = False
             if policy_key not in policies_used:
-                LOG.debug('removing orphaned policy {}'.format(policy_key))
+                LOG.debug("policy '{}' no longer referenced by a listener: "
+                          "({})".format(policy_key, policies_used))
                 self.lbdriver.purge_orphaned_l7_policy(
                     tenant_id=policy['tenant_id'],
                     l7_policy_id=policy_key,
-                    hostname=policy['hostnames'])
+                    hostnames=policy['hostnames'])
+                purged = True
+            elif not has_l7policies.get(policy['id'], False):
+                # should always be present on Neutron DB!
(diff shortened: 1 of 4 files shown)


ssorenso:
Re-Assigning to @jlongstaf as @richbrowne is unavailable today.

PR #1272, opened 2018-03-01
Fixing the agent code to make it compatible with Pike release.

kush1093:
@jlongstaf 
Fixes #<issueid>
WIP #<issueid>
...

#### What's this change do?
Import appropriate libraries to get agent/s working in Pike test environment.
From the top of the modified code.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent.py
@@ -33,9 +33,9 @@ except ImportError as CriticalError:
 
 
 try:
-    from neutron.agent.common import config
     from neutron.agent.linux import interface
     from neutron.common import config as common_config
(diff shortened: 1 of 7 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/agent.py
```diff
@@ -33,10 +33,10 @@
 
 
 try:
-    from neutron.agent.common import config
     from neutron.agent.linux import interface
     from neutron.common import config as common_config
     from neutron.common import rpc as n_rpc
+    from neutron.conf.agent import common as config
```
jlongstaf: I'm wondering if this code change will run in Ocata and Newton. We should test this change in Ocata and Newton before we commit, if you haven't already. You might need to put the import into a try/except block.

PR #1280, opened 2018-03-14
Remove references to internal F5 sites.

jlongstaf:
@szakeri 

#### What's this change do?
Because this code lives in a public repo, we need to remove all references to internal F5 sites. Removed two references, one a requirements file, and the other a doc string comment.

requirements.functest.txt

#### Any background context?
This is a requirement from F5 IT security. Users who run func tests must remember to install pytest symbols and not rely on the requirements file to do so. This happens automatically as part of testenv func tests setup.


szakeri:
Looks great

PR #1295, opened 2018-03-28
Provide IPv6 mask for IPv6 virtual servers.

### LINKED ISSUE #1294: Agent attempts to pass IPv4 netmask for IPv6 virtual server
#### Agent Version
This was encountered on agent version 10.2.0

#### Operating System
RHEL 7.4

#### OpenStack Release
Newton

#### Description
When attempting to create a Neutron LB on an IPv6 subnet, the agent passes the netmask as '255.255.255.255' to the LTM. This results in an HTTP 400 from the LTM due to the IP version mismatch and a failure to create the LB.

After making a small code modification, which will be in a PR shortly, the mask is passed as "ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff" and the LB is able to be created normally.

#### Deployment
The deployment under test is a standard HA Red Hat OpenStack 10 deployment with three controller nodes. Each controller node is running the F5 agent with a shared agent ID and managing the same standalone F5 VE appliance.

marbindrakon:
The VIP netmask was always given as 255.255.255.255 which caused errors
when placing a Neutron LB on an IPv6 subnet. This passes the proper /128
mask for IPv6 virtual server IP addresses.

@richbrowne 
Fixes #1294 
...

#### What's this change do?
This change causes the agent to pass "ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff" as the netmask for IPv6 VIPs rather than the previous value of 255.255.255.255. I did not do testing with IPv6 pool members.

service_adapter.py

#### Any background context?
This resolves a failure to provision a Neutron LB on an IPv6 subnet. In the test scenario, the LB is placed on a shared IPv6 subnet while the backends are on an IPv4 tenant network.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
@@ -451,8 +451,11 @@ class ServiceModelAdapter(object):
                 vip['destination'] = ip_address + ":" + str(port)
         else:
             LOG.error("No VIP address or port specified")
-
-        vip["mask"] = '255.255.255.255'
+        
+        if ':' in ip_address:
+            vip["mask"] = 'FFFF:FFFF:FFFF:FFFF:FFFF:FFFF:FFFF:FFFF'
+        else:
+            vip["mask"] = '255.255.255.255'
 
         if "admin_state_up" in listener:
             if listener["admin_state_up"]:


marbindrakon:
Can somebody re-run the CI for this commit? I don't see any failures related to the changed code.

hlesesne:
Bump.

marbindrakon:
Closing this PR as it is 3 years old now

PR #1302, opened 2018-04-03
Changing default logging option in agent INI

### LINKED ISSUE #1301: Change default logging option in agent INI
#### OpenStack Release
Newton

#### Description
The agent logging configuration is set to debug=True by default. The default should be changed to False. Also, update f5-sdk ManagementRoot object instantiation to set the debug flag based on how the setting of the agent INI (set to True only if debug=True in the agent INI file). See this documentation for more detail: https://f5-sdk.readthedocs.io/en/latest/userguide/debugging_api_calls.html. Note that starting with release 3.0.11, the default is set to turn off debug tracing in the SDK. https://github.com/F5Networks/f5-common-python/releases/tag/v3.0.11.

kush1093:
Issues: Default logging configuration option was set to True
Fixes #1301 

Problem: Default logging configuration option was set to True

Analysis: Issue is solved by changing default logging option to False

Tests: No tests

@jlongstaf 
#### What issues does this address? Changing the default logging option
Fixes #1301 
WIP #<issueid>
...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
@@ -293,6 +293,10 @@ OPTS = [  # XXX maybe we should make this a dictionary
         'trace_service_requests',
         default=False,
         help='Log service object.'
+    ),
+    cfg.BoolOpt(
+        'debug', default=False,
+        help='Enable debug option'
     )
 ]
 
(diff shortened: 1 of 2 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
```diff
@@ -665,7 +665,8 @@ def _open_bigip(self, hostname):
             bigip = ManagementRoot(hostname,
                                    self.conf.icontrol_username,
                                    self.conf.icontrol_password,
-                                   timeout=f5const.DEVICE_CONNECTION_TIMEOUT)
+                                   timeout=f5const.DEVICE_CONNECTION_TIMEOUT,
+                                   debug=True)
```
jlongstaf: We should only set debug=True when the agent INI setting is debug=True. To do so, add a config option to the OPT array in this file, something like:
cfg.BoolOpt(
        'debug',
        default=False,
        help='Enable debug logging.'
    )

You can then pass that value into the ManagementRoot constructor:
...
debug=self.conf.debug)

PR #1338, opened 2018-10-08

Niklaus-xie:
otherwise, seems when deleting rd with name like
Project_21c2550efe324765822917bda7abd3b5_aux_X, if
we replace rd name with partition, it would lead to:
Not Found for uri
net/route-domain/~Project_21c2550efe324765822917bda7abd3
b5~Project_21c2550efe324765822917bda7abd3b5


Niklaus-xie:
took a little look at the issue https://tron.f5net.com/sr/1-4448321425/#1-21KEYWX

adding Jeff for a review. I am not too familiar with the rd logic. But this logic
seems reasonable to me. Submitting this code change here for discussion and
merge if possible.

PR #1341, opened 2018-10-10

Niklaus-xie:
otherwise, seems when deleting rd with name like
Project_21c2550efe324765822917bda7abd3b5_aux_X, if
we replace rd name with partition, it would lead to:
Not Found for uri
net/route-domain/~Project_21c2550efe324765822917bda7abd3
b5~Project_21c2550efe324765822917bda7abd3b5

(cherry picked from commit c0c30ae079eae5c65e2ed251d2991322aeb3f2f1)


Niklaus-xie:
this was submitted before the reset, and it brings the ssl offloading patches which are not needed any more in this branch. will re-submit one.

PR #1342, opened 2018-10-10
Update version to 9.6.6

Niklaus-xie:
Update version to 9.6.6


Niklaus-xie:
don't merge this until other patches are merged/reverted later.

Niklaus-xie:
this was submitted before the reset, and it brings the ssl offloading patches which are not needed any more in this branch. will re-submit one.

PR #1352, opened 2018-10-18
9.7 release code change

zhang-shengping:
@zhaoqin-github 

WIP 9.7 release code change

This patch try to enable 9.7 release on github


Niklaus-xie:
code looks good to me . but it's probably better to leave the commit which modifies __init__.py as a separate pull request, and only merge that one until the very last minute.(in case we have any more commits at all)

PR #1373, opened 2018-11-19
expose valid esd names via agent

Niklaus-xie:
expose valid esd names via agent


Niklaus-xie:
a new one is submitted to get rid of unnecessary commit history

PR #1378, opened 2018-11-23
Hw 9.6.0 agen master

Niklaus-xie:
for review & discussion


Niklaus-xie:
seems unnecessary commit history exists. will clean those later

PR #1381, opened 2018-11-27
for the v6 case, we should ignore the netmask

jikui:
#### What's this change do?
For the v6 host  address, we could just ignore the netmask instead of using the fix "255.255.255.255“ which is for v4 case.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
@@ -452,7 +452,15 @@ class ServiceModelAdapter(object):
         else:
             LOG.error("No VIP address or port specified")
 
-        vip["mask"] = '255.255.255.255'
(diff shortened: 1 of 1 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/service_adapter.py
```diff
@@ -452,7 +452,15 @@ def _map_virtual(self, loadbalancer, listener, pool=None, policies=None):
         else:
             LOG.error("No VIP address or port specified")
 
-        vip["mask"] = '255.255.255.255'
+        # for the v6 ip address, do not explicitly assign the mask
+        try:
+            socket.inet_pton(socket.AF_INET,ip_address)
```
Niklaus-xie: socket has not been imported.

zhaoqin-github:
@jikui If 928ad12 is not required anymore, we may 'amend' that commit, instead of merge it.

PR #1387, opened 2018-12-10
Rd creation with same

zongzw:
PROBLEM
After testing new driver for HPB in our lab and technical environment, we just deployed the packages in Production. Unfortunately we are not able to create lbaas objects in internal networks. After some tests we discovered the problem is f5 agent is trying to create self ips using the same route domain id in both f5 units , however ids are assgined locally and they are different in both units, and therefore self ip can't be created in both units using the same route domain ids.

#### What's this change do?

Currently the f5-openstack-agent creates route domains in two cases:

- Whenever a tenant is created (tenants.py -> assure_tenant_create 74:87) 

- Whenever a service request is made for a tenant where the neutron subnet ID associated with a loadbalancer or pool member overlaps the CIDR block for a previously requests subnet ID (network_service -> _create_aux_rd 349:362)

Both of these cases interate through each bigip associated with the agent calling the network_helper.py -> create_route_domain method. This method in turn calls the _get_next_domain_id method which finds the lowest available route domain ID on that specific bigip device. This method does not check that the route domain ID returned by a particular bigip is consistant with the other bigips associated wtih the agent. If for some reason, the ids are not consistant, this creates a provisioning (create and delete) error. 

The recommendation is that the calls to network_helper.py create_route_domain and _get_next_domain_id be altered to include all bigips, removing the iteration. Within the _get_next_domain_id method, the available route domain IDs are collect from all associated bigip devices, the lowest route domain which is coherent on all devices selected and returned. The recommendation is to only return a route domain which is consistant on all bigips.

#### Initial test result in local sandbox: works ok: 
2018-12-10 17:36:50.457 4659 DEBUG f5_openstack_agent.lbaasv2.drivers.bigip.network_service [req-bbace356-67c4-41d1-be00-bb3ac09decf9 2d26c96aa0f345eaafc3f5b50d2bbd8e fde45211da0a44ecbf38cb0b644ab30d - - -] Allocated route domain 3 for tenant fde45211da0a44ecbf38cb0b644ab30d _create_aux_rd /usr/lib/python2.7/site-packages/f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py:377
2018-12-10 17:36:50.457 4659 DEBUG f5_openstack_agent.lbaasv2.drivers.bigip.network_service [req-bbace356-67c4-41d1-be00-bb3ac09decf9 2d26c96aa0f345eaafc3f5b50d2bbd8e fde45211da0a44ecbf38cb0b644ab30d - - -] Tenant fde45211da0a44ecbf38cb0b644ab30d now has 1 route domains assign_route_domain /usr/lib/python2.7/site-packages/f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py:345

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
@@ -73,13 +73,14 @@ class BigipTenantManager(object):
         # create tenant route domain
         if self.conf.use_namespaces:
             bigips = self.driver.get_all_bigips()
+            rd_id = self.network_helper.get_next_domain_id(bigips)
             for bigip in bigips:
                 if not self.network_helper.route_domain_exists(bigip,
                                                                folder_name):
                     try:
                         self.network_helper.create_route_domain(
                             bigip,
-                            bigips,
+                            rd_id,
                             folder_name,
                             self.conf.f5_route_domain_strictness)
                     except Exception as err:
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
@@ -72,12 +72,14 @@ class BigipTenantManager(object):
 
         # create tenant route domain
         if self.conf.use_namespaces:
-            for bigip in self.driver.get_all_bigips():
+            bigips = self.driver.get_all_bigips()
+            for bigip in bigips:
                 if not self.network_helper.route_domain_exists(bigip,
                                                                folder_name):
                     try:
                         self.network_helper.create_route_domain(
                             bigip,
+                            bigips,
                             folder_name,
                             self.conf.f5_route_domain_strictness)
                     except Exception as err:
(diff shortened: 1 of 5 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
```diff
@@ -72,12 +72,14 @@ def assure_tenant_created(self, service):
 
         # create tenant route domain
         if self.conf.use_namespaces:
-            for bigip in self.driver.get_all_bigips():
+            bigips = self.driver.get_all_bigips()
+            for bigip in bigips:
```
zhaoqin-github: I think my original idea is to pre-allocate the "next route domain id" before this loop. This implementation will still call create_route_domain() in this loop. Not sure if it can fix the problem either.

zongzw: Right, that would be still the problem as doing like this...

zongzw:
close it and reopen a new pull request for refactor commits. 

https://github.com/F5Networks/f5-openstack-agent/pull/1388

PR #1391, opened 2018-12-19
performance issue fix: only do action on pending listeners.

zongzw:
When Creating / Deleting listeners, all the listeners would be taken action repeatly. 
In fact, only pending status should be take care of.

Moreover, pool, member, healthmonitor are also be the same situation. 
Test result: 

default

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py
@@ -179,6 +179,13 @@ class LBaaSBuilder(object):
         bigips = self.driver.get_config_bigips()
 
         for listener in listeners:
+            # only take action on pending_xx status listeners.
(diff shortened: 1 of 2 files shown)


zhang-shengping:
Hello, Andrew I checked your code, I found, in fact, we did the same thing at begining.

As I code:
```
         service["listeners"] = [listener]
         self._common_service_handler(service)
```
I put the target listener into service["listeners"], therefore there is one "PENDING_CREATE"  listener in service["listeners"] that pass to the assure_service method. 

Indeed, I missing the point of the slow creating listener (pool, member) problem by mistake thinking of RPC call problem. As you said, in fact, long pending creation is caused by serialized icontrol calls.

zongzw:
@zhang-shengping I didn't think about the 'second bigip device' situation you mention. need more test. thanks. 
What's more, we need to check if there's any other code legic which causes time waste in the code. 
RPC is still worth to suspect. .

PR #1405, opened 2019-02-11
F5 P/O agent code change for Openstack Queens lbaasv2

zhang-shengping:
1. change code for Openstack Queens version
2. fix some old bugs

(cherry picked from commit 9d56e482b7045bb0d0036b25f33144ab9cf61906)

Conflicts:
	f5_openstack_agent/lbaasv2/drivers/bigip/lbaas_builder.py

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent.py
@@ -33,9 +33,9 @@ except ImportError as CriticalError:
 
 
 try:
-    from neutron.agent.linux import interface
     from neutron.common import config as common_config
     from neutron.common import rpc as n_rpc
+    from neutron.conf.agent.common import INTERFACE_OPTS
     try:
         from neutron.conf.agent import common as config
     except Exception:
@@ -71,7 +71,7 @@ def main():
     """F5 LBaaS agent for OpenStack."""
     cfg.CONF.register_opts(OPTS)
     cfg.CONF.register_opts(manager.OPTS)
-    cfg.CONF.register_opts(interface.OPTS)
+    cfg.CONF.register_opts(INTERFACE_OPTS)
 
     config.register_agent_state_opts_helper(cfg.CONF)
     config.register_root_helper(cfg.CONF)
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/esd_filehandler.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/esd_filehandler.py
@@ -50,8 +50,8 @@ class EsdJSONValidation(object):
                     self.esdJSONDict.update(fileJSONDict)
 
             except ValueError as err:
-                    LOG.error('ESD JSON File is invalid: %s', err)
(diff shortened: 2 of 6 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/test/test_vcmp.py
```diff
@@ -126,9 +126,9 @@ def old_setup_vcmp_method_test(mock_mgmt_root, mock_driver, mock_host_obj):
 def test___init__(setup_vcmp_mgr):
     vcmp = setup_vcmp_mgr
     assert vcmp.vcmp_hosts != []
-    assert len(vcmp.vcmp_hosts) is 2
+    assert len(vcmp.vcmp_hosts) == 2
```
Niklaus-xie: these are common places which IMO is irrelevant to queens_engineering. put them into a separate commit so that they are submitted into each branch we need?(e.g. 9.6 9.7 etc, if needed).

f5_openstack_agent/lbaasv2/drivers/bigip/agent.py
```diff
@@ -71,7 +71,7 @@ def main():
     """F5 LBaaS agent for OpenStack."""
     cfg.CONF.register_opts(OPTS)
     cfg.CONF.register_opts(manager.OPTS)
-    cfg.CONF.register_opts(interface.OPTS)
+    cfg.CONF.register_opts(INTERFACE_OPTS)
```
Niklaus-xie: this way seems to make it quite complicated. should probably Try this way: copy the content of  INTERFACE_OPTS  into this file , then  this line can be always the same between each branch.  each branch has the same and we dont have to consider queens differently.

f5_openstack_agent/lbaasv2/drivers/bigip/esd_filehandler.py
```diff
@@ -50,8 +50,8 @@ def read_json(self):
                     self.esdJSONDict.update(fileJSONDict)
 
             except ValueError as err:
-                    LOG.error('ESD JSON File is invalid: %s', err)
-                    raise f5_ex.esdJSONFileInvalidException()
+                LOG.error('ESD JSON File is invalid: %s', err)
```
Niklaus-xie: same as above.

f5_openstack_agent/lbaasv2/drivers/bigip/test/test_vcmp.py
zhang-shengping: I will do it, after I fix the loopback4 things, or you can help me a bit :D

f5_openstack_agent/lbaasv2/drivers/bigip/agent.py
Niklaus-xie: in that way, we do not need a separate queens branch.   we can still use our existing 9.8 branch for queens, as well as the original M N, O. thus,  avoid many inconveniences for maintenance work.

zhang-shengping: agree

PR #1424, opened 2019-06-18
Refresh esd 9.8 with trigger

zhang-shengping:
signal to trigger ESD refresh

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
@@ -498,6 +498,12 @@ class LbaasAgentManager(periodic_task.PeriodicTasks):  # b --> B
             self.lbdriver.connect()
 
     @periodic_task.periodic_task(spacing=PERIODIC_TASK_INTERVAL)
+    def refresh_esd(self, context):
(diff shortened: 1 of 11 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
```diff
@@ -134,6 +134,11 @@
         'password_cipher_mode',
         default=False,
         help='The flag indicating the password is plain text or not.'
+    ),
+    cfg.BoolOpt(
+        'esd_auto_refresh',
+        default=False,
```
zhang-shengping: change to True

f5_openstack_agent/lbaasv2/drivers/bigip/refersh_esd.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/python
```
zongzw: Can we move this file to <repo>/bin?

zhang-shengping: Yes, it will be installed to /usr/bin/ while rpm is installing.
You can refer to the setup.py file.

PR #1439, opened 2019-10-11
multi-provider with multi-snat-pool-addresses

zhang-shengping:
enable  multi-provider with multi-snat-pool-addresses

### DIFF
--- a/etc/neutron/services/f5/f5-openstack-agent.ini
+++ b/etc/neutron/services/f5/f5-openstack-agent.ini
@@ -24,7 +24,7 @@
 #             ####| |#########| |###
 #             ####| |#########| |##
 #              ###| |########/ /##
-#               #|    |####|  /## 
+#               #|    |####|  /##
 #                ##############
 #                 ###########
 #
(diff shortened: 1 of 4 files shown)


etc/neutron/services/f5/f5-openstack-agent.ini
```diff
@@ -409,35 +409,44 @@ f5_route_domain_strictness = False
...
 # addresses per active traffic-group at the time
 # a service is provisioned.
 #
-# This setting will be forced to 0 (zero) if 
+# This setting will be forced to 0 (zero) if
 # f5_global_routed_mode = True.
 #
 f5_snat_addresses_per_subnet = 1
+
+# This provider_name use to distinguish different
```
zongzw: Suppose the provider_name can have more usages in the future because it's an identifier of provider/plugin. 
So how about to say: 

```
# This provider_name is the identifier of f5driver. Currently it is used to distinguish ... 
```

Not sure of that, you decide.

zongzw (approved):
lgtm.

PR #1454, opened 2020-05-15
support for route domain

jikui:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

#### What's this change do?
change for the route domain configure.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
@@ -54,6 +54,12 @@ class NetworkHelper(object):
         'floodingType': 'none',
     }
 
+    route_domain_update = {
+        'name': None,
+        'partition': None,
+        'bwcPolicy': None,
+    }
+
     route_domain_defaults = {
         'name': None,
         'partition': '/' + const.DEFAULT_PARTITION,
@@ -299,7 +305,7 @@ class NetworkHelper(object):
         return lowest_available_index
 
     @log_helpers.log_method_call
-    def create_route_domain(self, bigip, rd_id,
+    def create_route_domain(self, bigip, rd_id, qos,
                             partition=const.DEFAULT_PARTITION,
                             strictness=False, is_aux=False, name=None):
         """Creates the route domain based upon settings in config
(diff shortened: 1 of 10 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
```diff
@@ -34,6 +35,8 @@ def __init__(self, conf, driver):  # XXX maybe we need a better name: conf
         self.system_helper = SystemHelper()
         self.network_helper = NetworkHelper()
         self.service_adapter = self.driver.service_adapter
+        self.rd_helper = resource_helper.BigIPResourceHelper(
+            resource_helper.ResourceType.route_domain)
 
     def assure_tenant_created(self, service):
```
zhang-shengping, line 41: did you test delete for a tenant, for example, a tenant without any resource inside on BIGIP, it will be cleanup by F5 agent. Is it will be successful as well. 

what is the corresponding OpenStack QoS bonded resource? is it will be clean up with the last loadbalancer in a tenant (partition), the partition could be clean up successfully in BIGIP  as well.

zhang-shengping (approved):

f5_openstack_agent/lbaasv2/drivers/bigip/network_helper.py
```diff
@@ -328,6 +335,10 @@ def create_route_domain(self, bigip, rd_id,
         payload['name'] = name
         payload['partition'] = '/' + partition
         payload['id'] = rd_id
+        if len(qos) == 0:
+            payload['bwcPolicy'] = 'None'
```
zhaoqin-github, line 339: Not sure why we need to use 'None', instead of None. Any purpose?

f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
```diff
@@ -81,13 +84,30 @@ def assure_tenant_created(self, service):
...
                             "Failed to create route domain for "
                             "tenant in %s" % (folder_name))
+                else:
+                    LOG.debug("Try to update route domain")
+                    try:
+                        payload = self.network_helper.route_domain_update
+                        payload['name'] = folder_name
+                        payload['partition'] = folder_name
+                        if service['qos'] == '':
+                            payload['bwcPolicy'] = 'None'
```
zhaoqin-github, line 102: Not sure why we need to use 'None', instead of None. Any purpose?

jikui: Qin, the "None" is the fixed profile name at bigip side which means no any bwc control profile is bound to bigip. 
We had better keep this as "None" in order to keep the consistence between openstack and bigip.

zhaoqin-github (commented):
I feel basically OK with the code, but please resolve code conflict. Thanks!

PR #1491, opened 2020-10-22
Fix: check vips left on bigip rather than vs.

zhang-shengping:
Before we delete SNAT port and VIP under certain PENDING_DELETE LB,
we need to check if the subnet of the lB is being used in the
route domain of the LB.

To do this, we check there is any VIPs or Nodes are associated with
the route domain and subnet of the LB.


zongzw (approved):
LGTM

zhang-shengping:
> LGTM

Thanks

PR #1492, opened 2020-10-22
Fix: check any VIPs and Nodes left before delete a partition.

zhang-shengping:
If there is nothing (VIPs and Nodes) left on a partition,
then we can delete that partition.

Do not let the user add a node or a vip manually by web GUI.


zongzw (approved):
LGTM.

zhang-shengping:
> LGTM.

Thanks

PR #1503, opened 2020-10-26
pool descrition could not be updated at bigip side

jikui:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


zhaoqin-github (changes requested):
If lbaas pool description never impact any bigip configuration, we may not need to add this support.

jikui:
> If lbaas pool description never impact any bigip configuration, we may not need to add this support.

Agree with you. This issue also exists in the old logic.

PR #1510, opened 2020-10-28
pool's loadbalancerMode might be changed by member operations

jikui:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


zhaoqin-github (changes requested):
Should we also adjust pool's loadbalancing mode, when deleting members?

jikui:
> Should we also adjust pool's loadbalancing mode, when deleting members?

yes, added the logic into delete method as well.

zhaoqin-github (approved):

PR #1518, opened 2020-11-04
hotfix: delete previous code change for deleting last one loadbalancer in agent 2.0

zhang-shengping:
Previous delete last one loadbalancer, the partition needs to be purged,
agent 2.0 has an approach to solve the problem by finding the last one lb in neutron DB.

However, we found another approach to solve the problem in agent 3.0 and
2.0.

I forget to delete the previous code change in f5 agent 2.0, and test. and it causes a bug.


zhaoqin-github:
Won't require this change.

PR #1523, opened 2020-11-09
Add TERMINATED_HTTPS protocol type vs support

zongzw:
Tested with multiple creation and deletion. 

Notice that, the certificate file would be left under /Common (System -> Certificate Managerment). 
It's by design or bug, however, it should be another story.


zhaoqin-github (approved):
Approve this commit. Two suggestions for the next commit:

1. LBaaS listener api allows to update tls container ref and sni container ref. Please help to confirm that and implement.

2. If possible, agent lite should allow user to manually modify the settings in client ssl profile (may not include its reference to keys uploaded by agent). Please help to verify if the manual modifications in client ssl profile will not be impact after updating profile.

zongzw:
> Approve this commit. Two suggestions for the next commit:
> 
>     1. LBaaS listener api allows to update tls container ref and sni container ref. Please help to confirm that and implement.
> 
I didn't find the very link showing the detail of lbaas v2 listener update api. Please paste it here, thanks. 
See from neutron command line, there is no parameter to let users updating tls or sni refs in 'neutron lbaas-listener-update'
 

>     2. If possible, agent lite should allow user to manually modify the settings in client ssl profile (may not include its reference to keys uploaded by agent). Please help to verify if the manual modifications in client ssl profile will not be impact after updating profile.

The current behavior is, when user modify any field of the listener, e.g listener's name, the client ssl profile would be restored to original one. That means lbaas-listener-update will overwrite the manual client ssl configuration on BIG-IP.

Will all the profile settings(like persistence, http) support manual configuration in future? Till now I didn't see the supportibility for those profiles.

zhaoqin-github:
> > Approve this commit. Two suggestions for the next commit:
> > ```
> > 1. LBaaS listener api allows to update tls container ref and sni container ref. Please help to confirm that and implement.
> > ```
> 
> I didn't find the very link showing the detail of lbaas v2 listener update api. Please paste it here, thanks.
> See from neutron command line, there is no parameter to let users updating tls or sni refs in 'neutron lbaas-listener-update'
> 
> > ```
> > 2. If possible, agent lite should allow user to manually modify the settings in client ssl profile (may not include its reference to keys uploaded by agent). Please help to verify if the manual modifications in client ssl profile will not be impact after updating profile.
> > ```
> 
> The current behavior is, when user modify any field of the listener, e.g listener's name, the client ssl profile would be restored to original one. That means lbaas-listener-update will overwrite the manual client ssl configuration on BIG-IP.
> 
> Will all the profile settings(like persistence, http) support manual configuration in future? Till now I didn't see the supportibility for those profiles.

1. Please read https://github.com/openstack/neutron-lbaas/blob/stable/pike/neutron_lbaas/extensions/loadbalancerv2.py#L221

2. Agent lite need to support manual configuration against all possible profiles. That is the core value of Agent lite. Persistence and http profile already support that. Client ssl profile also need to support that right away.

PR #1541, opened 2020-11-20
fix: VS is not bond to correct VIP in L2 mode

zhang-shengping:
BUG: VS is bond to a VIP that is not with a route domain in L2 mode

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
@@ -523,6 +523,14 @@ class ListenerManager(ResourceManager):
     @serialized('ListenerManager.create')
     @log_helpers.log_method_call
     def create(self, listener, service, **kwargs):
+        loadbalancer = service.get("loadbalancer", None)
+        traffic_group = self.driver.service_to_traffic_group(service)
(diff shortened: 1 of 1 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -534,6 +542,15 @@ def update(self, old_listener, listener, service, **kwargs):
     @serialized('ListenerManager.delete')
     @log_helpers.log_method_call
     def delete(self, listener, service, **kwargs):
+        loadbalancer = service.get("loadbalancer", None)
```
zhaoqin-github: What is the purpose to run this piece of code before deleting a VS? The resource path of a VS does not contain route domain information. What is your expected result?

zhaoqin-github (changes requested):

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
zhang-shengping: > What is the purpose to run this piece of code before deleting a VS? The resource path of a VS does not contain route domain information. What is your expected result?

Yes, do not need this, vs is deleted by name, not whole payload.

zhaoqin-github (approved):

PR #1544, opened 2020-11-22
Add supporting to update listener with tls changes.

zongzw:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
@@ -329,6 +329,17 @@ class ListenerManager(ResourceManager):
         if old_listener['default_pool_id'] != listener['default_pool_id']:
             payload['persist'] = create_payload['persist']
 
+        if listener['protocol'] == "TERMINATED_HTTPS":
+            LOG.debug("Finding tls setting  differences.")
+            old_sni = [n['tls_container_id']
+                       for n in old_listener['sni_containers']]
+            sni = [n['tls_container_id']
+                   for n in listener['sni_containers']]
+            if old_listener['default_tls_container_id'] != \
+               listener['default_tls_container_id'] or old_sni != sni:
+                payload['tls'] = "tls change"
+                LOG.debug("tls changes happen")
+
         return super(ListenerManager, self)._update_payload(
             old_listener, listener, service,
             payload=payload, create_payload=create_payload
@@ -498,6 +509,25 @@ class ListenerManager(ResourceManager):
         super(ListenerManager, self)._create(bigip, vs, listener, service)
 
     def _update(self, bigip, vs, old_listener, listener, service):
+        # Add conditions here for more update requests via vs['profiles']
+        if 'tls' in vs:
+            vs_manager = resource_helper.BigIPResourceHelper(
+                resource_helper.ResourceType.virtual)
+            l_objs = vs_manager.get_resources(
+                bigip, partition=vs['partition'], expand_subcollections=True)
+            v = filter(lambda x: x.name == vs['name'], l_objs)[0]
+            orig_profiles = v.profilesReference
+
+            if 'profiles' not in vs:
+                vs['profiles'] = list()
+            vs['profiles'] += orig_profiles['items']
+            for n in vs['profiles']:
+                if n['context'] == 'clientside':
+                    vs['profiles'].remove(n)
+            tls = self.driver.service_adapter.get_tls(service)
+            if tls:
+                self._create_ssl_profile(bigip, vs, tls)
+
         persist = None
         if vs.get('persist'):
             persist = service[self._key]['session_persistence']
@@ -509,6 +539,9 @@ class ListenerManager(ResourceManager):
 
         super(ListenerManager, self)._update(bigip, vs, old_listener, listener,
                                              service)
+
+        old_service = {"listener": old_listener}
+        self._delete_ssl_profiles(bigip, vs, old_service)
 
     def _delete(self, bigip, vs, listener, service):
         super(ListenerManager, self)._delete(bigip, vs, listener, service)


f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -499,6 +519,9 @@ def _create(self, bigip, vs, listener, service):
         super(ListenerManager, self)._create(bigip, vs, listener, service)
 
     def _update(self, bigip, vs, old_listener, listener, service):
+        tls = self.driver.service_adapter.get_tls(service)
+        if tls:
+            self._create_ssl_profile(bigip, vs, tls)
```
zhaoqin-github: If the user manually modify some settings in client ssl profile, will your code keep those unchanged and only modify the certificates?

zhaoqin-github: And if the VS also has some other profiles manually inserted by the user, will those profiles still attach to the VS after updating certificate?

zhaoqin-github (changes requested):

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
zongzw: > If the user manually modify some settings in client ssl profile, will your code keep those unchanged and only modify the certificates?

Yes, this  code will  keep the manual-configuration in that clientssl  profile  unchanged.

zongzw: > And if the VS also has some other profiles manually inserted by the user, will those profiles still attach to the VS after updating certificate?

No for that.
Users  insert other profiles like 'http  profile' manually, even only vs' name is  changed( no clientssl  proflie changes), the http  profile will be lost..  
Is this a bigger  story  that our  _update  function  should be considered  carefullly?

zhaoqin-github: I see this piece of code will add "profiles" list to vs payload before updating, so that the existing profiles of the vs will be removed, only client ssl profile will be there. Will that happen?

https://github.com/F5Networks/f5-openstack-agent/blob/9.8-stable/f5_openstack_agent/lbaasv2/drivers/bigip/listener_service.py#L220

zhaoqin-github (changes requested):

zongzw (commented):
That's right. It would happen.

Does that mean any existing profiles in `vip['profiles']` need to be pre-loaded via `bigip.tm.ltm.virtuals.virtual.load` and merged with `payload` in order to keep unchanged after update?

zhaoqin-github:
> That's right. It would happen.
> 
> Does that mean any existing profiles in `vip['profiles']` need to be pre-loaded via `bigip.tm.ltm.virtuals.virtual.load` and merged with `payload` in order to keep unchanged after update?

I think usually _update() should never modify vs['profiles']. I can only patch VS properties, which are not a sub-collection.
If it does need to attach a new profile to VS, it should invoke sub-collection's api to attach a new profile to VS. It is something like this:
https://github.com/F5Networks/f5-openstack-agent/blob/9.8-stable/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py#L1113

Personally, I think vs['profiles'] is not a very good way to attach/detach profiles. It may only used when creating a new VS.

In you case, I feel we may assume the client ssl profile has already been attached to the VS, so that we only need to modify the cert files and patch the content of client ssl profile itself. Does that work?

zongzw:
> > That's right. It would happen.
> > Does that mean any existing profiles in `vip['profiles']` need to be pre-loaded via `bigip.tm.ltm.virtuals.virtual.load` and merged with `payload` in order to keep unchanged after update?
> 
> I think usually _update() should never modify vs['profiles']. I can only patch VS properties, which are not a sub-collection.
> If it does need to attach a new profile to VS, it should invoke sub-collection's api to attach a new profile to VS. It is something like this:
> https://github.com/F5Networks/f5-openstack-agent/blob/9.8-stable/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py#L1113
> 
> Personally, I think vs['profiles'] is not a very good way to attach/detach profiles. It may only used when creating a new VS.
> 
> In you case, I feel we may assume the client ssl profile has already been attached to the VS, so that we only need to modify the cert files and patch the content of client ssl profile itself. Does that work?

In your suggestion, it still cannot handle the "adding one more SNI container id" case, i.e. listener X has default_tls_container_id A, and sni_container_refs [B], now we want to add one more sni_container_refs to [B C].

How to bind a profile to a vs in resource_helper? Do we have any existing profile handling logic for reference, like HTTP / oneconnect / persistence profiles? I have no idea how to handle profile type without `vip['profiles']' help.

In my personal thinking, using `vip['profiles']` can leverage much of the original logic in the legacy agent. What's more, code logic related to cert management is not so simple as policies. It has a separate encapsulation `ssl_profile.py`, It seems a big part of logic. 

The SSL resources are created under 'Common', which is also different from policies. And the resource naming keeps the same with container id in the barbican system. It would make chaos if client SSL A on BIG-IP has B's content in barbican, I think.

zhaoqin-github:
> https://github.com/F5Networks/f5-openstack-agent/blob/9.8-stable/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py#L1113

I did not see ssl profile naming code. If we do not have a unique ssl profile for each VS, you may keep its existing way and carefully use vip['profiles'] to update VS profiles. Maybe we would load all its attached profile names before patching, in order to avoid detaching user added profiles.

zhaoqin-github (approved):
Suggest to add a case in pipeline to ensure ssl profile is update and other profiles will not be impacted.

zongzw:
> Suggest to add a case in pipeline to ensure ssl profile is update and other profiles will not be impacted.

Right, there is.

PR #1552, opened 2020-12-03
leave placeholder for bulk

Niklaus-xie:
and add a switch to pass different values for host.


Niklaus-xie:
since skip rh6 is merged in another pr,  i submitted new separate one and close this one. see: https://github.com/F5Networks/f5-openstack-agent/pull/1554 instead .

PR #1557, opened 2020-12-08
only add the new create http profile when it was created

jikui:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
@@ -305,12 +305,14 @@ class ListenerManager(ResourceManager):
             "default_pool_id": "pool",
             "connection_limit": "connectionLimit"
         }
+        self.http_profile_created = False
 
     def _create_payload(self, listener, service):
         payload = self.driver.service_adapter.get_virtual(service)
         profiles = payload.get('profiles', [])
-        if '/Common/http' in profiles:
-            profiles.remove('/Common/http')
+        if self.http_profile_created is True:
+            if '/Common/http' in profiles:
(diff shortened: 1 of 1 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -305,12 +305,14 @@ def __init__(self, driver):
...
             "connection_limit": "connectionLimit"
         }
+        self.http_profile_created = False
 
     def _create_payload(self, listener, service):
         payload = self.driver.service_adapter.get_virtual(service)
         profiles = payload.get('profiles', [])
-        if '/Common/http' in profiles:
-            profiles.remove('/Common/http')
+        if self.http_profile_created is True:
```
zongzw, line 313: LGTM. 

How about using `if self.driver.conf.f5_extended_profile:` here, so that we can make the listener creation fails directly when http profile fails, -- instead of listener creation succeed with wrong profile there (/Common/http). 

Both ways are ok to me. We may talk to decide.

zhaoqin-github (changes requested):
It is not a bug

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
jikui: andrew, even if self.driver.conf.f5_extended_profile is configured, if there is something wrong, such as the file doesn't exists , or the format is not correct, the http profile will not created either.  so, we had better use a brand new variant for this case.

zongzw: agree.

PR #1558, opened 2020-12-08
WIP: add check and sync data (need more tests)

zhang-shengping:
This feature has some limitations:

    1. If only vips are missing, the sync_data.py command will recover them.

    2. when vlan or selfip are missing, if we want to recover them,
       the associated vips need to be deleted.

    3. the route domain can not be missing, or it will cause route domain
       creation problem.

    4. This requires more test in case of affections of 2.0 agent, since
       some caches are deleted in some commonly used function in 2.0 agent.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
@@ -707,11 +707,16 @@ class iControlDriver(LBaaSBaseDriver):
             bigip.status_message = 'requesting iControl endpoint'
             LOG.info('opening iControl connection to %s @ %s' %
                      (self.conf.icontrol_username, hostname))
+#            bigip = ManagementRoot(hostname,
+#                                   self.conf.icontrol_username,
+#                                   self.conf.icontrol_password,
+#                                   timeout=f5const.DEVICE_CONNECTION_TIMEOUT,
+#                                   debug=self.conf.debug)
             bigip = ManagementRoot(hostname,
                                    self.conf.icontrol_username,
                                    self.conf.icontrol_password,
                                    timeout=f5const.DEVICE_CONNECTION_TIMEOUT,
-                                   debug=self.conf.debug)
+                                   debug=False)
             bigip.status = 'connected'
             bigip.status_message = 'connected to BIG-IP'
             self.__bigips[hostname] = bigip
@@ -1113,9 +1118,14 @@ class iControlDriver(LBaaSBaseDriver):
         """Set Agent Report State."""
         self.agent_report_state = report_state_callback
 
+    # pzhang: for agent 3.0
+    def neutron_service_exists(self, service, mgr=None):
+        return self._neutron_service_exists(service, mgr)
+
     def service_exists(self, service):
         return self._service_exists(service)
 
+    # pzhang 3.0 agent may not need these caches
     def flush_cache(self):
         # Remove cached objects so they can be created if necessary
         for bigip in self.get_all_bigips():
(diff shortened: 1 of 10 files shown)


zongzw (commented):
Can we redirect the signal handler to `def periodic_resync(self, context):`?

zhang-shengping:
> Can we redirect the signal handler to `def periodic_resync(self, context):`?

We could do that, but we still need to separate 2.0 and 3.0 code.

zongzw:
> > Can we redirect the signal handler to `def periodic_resync(self, context):`?
> 
> We could do that, but we still need to separate 2.0 and 3.0 code.

Right.

f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
```diff
@@ -707,11 +707,16 @@ def _open_bigip(self, hostname):
             bigip.status_message = 'requesting iControl endpoint'
             LOG.info('opening iControl connection to %s @ %s' %
                      (self.conf.icontrol_username, hostname))
+#            bigip = ManagementRoot(hostname,
+#                                   self.conf.icontrol_username,
+#                                   self.conf.icontrol_password,
+#                                   timeout=f5const.DEVICE_CONNECTION_TIMEOUT,
+#                                   debug=self.conf.debug)
```
zhang-shengping, line 714: here is not part of the code， I  will change it later

f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
```diff
@@ -587,15 +596,19 @@ def _assure_subnet_snats(self, assure_bigips, service, subnetinfo):
         snats_per_subnet = self.conf.f5_snat_addresses_per_subnet
         lb_id = service['loadbalancer']['id']
 
-        assure_bigips = \
+        temp_bigips = \
             [bigip for bigip in assure_bigips
                 if tenant_id not in bigip.assured_tenant_snat_subnets or
                 subnet['id'] not in
                 bigip.assured_tenant_snat_subnets[tenant_id]]
```
zhang-shengping, line 604: Here is not part of the code change, I will recover it later

f5_openstack_agent/lbaasv2/drivers/bigip/selfips.py
```diff
@@ -102,15 +103,19 @@ def assure_bigip_selfip(self, bigip, service, subnetinfo):
...
+        # pzhang on different bigip device
         # If we have already assured this subnet.. return.
         # Note this cache is periodically cleared in order to
         # force assurance that the configuration is present.
-        if tenant_id in bigip.assured_tenant_snat_subnets and \
-                subnet['id'] in bigip.assured_tenant_snat_subnets[tenant_id]:
-            return True
+        # if tenant_id in bigip.assured_tenant_snat_subnets and \
+        #         subnet['id'] in bigip.assured_tenant_snat_subnets[tenant_id]:
+        #     return True
```
zhang-shengping, line 118: need test, if this change will affect 2.0 agent

f5_openstack_agent/lbaasv2/drivers/bigip/l2_service.py
```diff
@@ -186,8 +186,9 @@ def assure_bigip_network(self, bigip, network):
                       'Attempted to assure a network with no id..skipping.')
             return
 
-        if network['id'] in bigip.assured_networks:
-            return
+        # pzhang: 3.0 and sync do not need cache here
```
zhang-shengping, line 189: need the test, if this change will affect 2.0 agent

PR #1563, opened 2020-12-11
fix lb operating_status for creating, updating and deleting loadbalancer

zhang-shengping:
fix lb operating_status for creating, updating, and deleting loadbalancer


zhaoqin-github (changes requested):
Do we really need to modify operating status, when updating or deleting?

zhang-shengping:
> Do we really need to modify operating status, when updating or deleting?

I think we need to keep consistency, If we update or deleting loadbalancer error, the operating status keeps ONLINE, does it seem right ?

PR #1575, opened 2020-12-23
add release note

zhang-shengping:


olvandeng (changes requested):
v9.8.21 (Pike, Queens) should be v9.9.0(Pike, Queens)

olvandeng (approved):

PR #1580, opened 2020-12-26
create snat_ip for backward compatibility

zhang-shengping:


zongzw (commented):
lgtm

zongzw (approved):

zongzw (approved):
Agree, ubuntu is no use. No need to build every time.

PR #1590, opened 2020-12-30
do not break member deletion

Niklaus-xie:
if the member is not found on devices.
otherwise, the member can not be deleted.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
@@ -1033,16 +1033,27 @@ class MemberManager(ResourceManager):
         bigips = self.driver.get_config_bigips()
         loadbalancer = service.get('loadbalancer')
         for bigip in bigips:
-            pool_resource = self._pool_mgr.pool_helper.load(
-                bigip,
-                name=urllib.quote(pool_payload['name']),
-                partition=pool_payload['partition']
-            )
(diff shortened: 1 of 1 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -1042,7 +1042,16 @@ def delete(self, resource, service, **kwargs):
...
-            member_resource.delete()
+            try:
+                member_resource.delete()
+            except HTTPError as err:
+                if err.response.status_code == 404:
+                    LOG.warning("the member not found, am ignoring")
+                    LOG.warning(str(err))
+                else:
+                    LOG.error("unknow member deletion error")
+                    raise err
```
Niklaus-xie: maybe not raise at all.?

jikui (approved):

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -1042,7 +1042,16 @@ def delete(self, resource, service, **kwargs):
                 name=urllib.quote(payload['name']),
```
zhaoqin-github: If the member does not exist, will 404 is raised when loading?

Niklaus-xie: looks like  including load of pool & member into the try block is fine.

PR #1615, opened 2021-01-08
Add avaliability_zone configuration for agent state reporting

zhang-shengping:

### DIFF
--- a/etc/neutron/services/f5/f5-openstack-agent.ini
+++ b/etc/neutron/services/f5/f5-openstack-agent.ini
@@ -446,7 +446,11 @@ f5_snat_addresses_per_subnet = 1
 # (Note)
 # This value must be configured with different values among
 # different neutron lbaasv2 providers.
-provider_name = 'f5networks'
+provider_name =
+
+# This configuration is used to report the availability_zone
+# to which the agent is belong.
+availability_zone =
 
 #
 # This setting will cause all networks to be
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager.py
@@ -94,6 +94,11 @@ OPTS = [
         help=('provider_name for snat pool addresses')
     ),
     cfg.StrOpt(
+        'availability_zone',
+        default=None,
+        help=('availability_zone for agent reporting')
+    ),
+    cfg.StrOpt(
         'agent_id',
         default=None,
         help=('static agent ID to use with Neutron')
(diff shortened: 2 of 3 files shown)


etc/neutron/services/f5/f5-openstack-agent.ini
```diff
@@ -446,7 +446,11 @@ f5_snat_addresses_per_subnet = 1
 # (Note)
 # This value must be configured with different values among
 # different neutron lbaasv2 providers.
-provider_name = 'f5networks'
+provider_name =
```
zongzw, line 449: Why do we remove the default value for `provider_name`? any concern?

zhang-shengping: just in case that the user configures it as f5networks by mistake.

zhang-shengping: > Why do we remove the default value for `provider_name`? any concern?

just in case the user configures provider_name as f5networks by mistake. 
let it empty to let users notice something miss, then configure it properly.

zongzw: what values should 'provider_name' be? why cannot users configure it as 'f5networks'?

zhang-shengping: > what values should 'provider_name' be? why cannot users configure it as 'f5networks'?

it is related to https://github.com/F5Networks/f5-openstack-agent/pull/1450/commits/059cde518429debf5878d601c2ecdb4efab4aef7

it is the case of multiple providers use the same subnet that should have a different snat pool for each provider.
In fact, it is not quite related to this case. I just change the provider_name by convenience.

zongzw: To my understanding, the provider_name will affect snat addresses' name. Making it empty seems no difference here. 
Letting different providers configure different provider_name may help them using different snat pool, that's the very usage, right?

We have 4 similar configurations now, may need to make clear, like agent_id, environment, provider_name, availability_zone.

zhang-shengping: > To my understanding, the provider_name will affect snat addresses' name. Making it empty seems no difference here.
> Letting different providers configure different provider_name may help them using different snat pool, that's the very usage, right?
> 
> We have 4 similar configurations now, may need to make clear, like agent_id, environment, provider_name, availability_zone.

Letting the provider_name empty will keep the agent as backward compatible by default if the user forgets to change it.

the 4 similar configurations are both pointed out in configuration comments of file f5-opentack-agent.ini and https://github.com/olvandeng/openstack-docs/blob/main/F5-%E9%85%8D%E7%BD%AE%E6%96%87%E6%A1%A3/F5-Agent-%E7%8E%B0%E7%BD%91%E5%B8%B8%E7%94%A8%E9%85%8D%E7%BD%AE.md .

If this commit pass, the availability_zone configuration will be added to both of the above files.

zongzw (approved):

PR #1646, opened 2021-01-31
use a new way to check if the tls or customized properties are changed or not.

jikui:
…for updating operation

@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


zhaoqin-github (changes requested):
Do not recommend to overwrite update() so a lot. It make the code become much more harder.

zhaoqin-github (approved):

PR #1649, opened 2021-02-03
Add server/client tcp profile

zhang-shengping:
1. support IPv6 and IPv4
2. support create/update/delete
2. support all upper-level protocol for TCP TOA

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/irule.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/irule.py
@@ -1,0 +1,234 @@
+# coding=utf-8
+# Copyright (c) 2016-2018, F5 Networks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from f5_openstack_agent.lbaasv2.drivers.bigip import resource_helper
+from oslo_log import log as logging
+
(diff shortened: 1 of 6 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/irule.py
```diff
@@ -0,0 +1,234 @@
...
+        self.vs_helper = resource_helper.BigIPResourceHelper(
+            resource_helper.ResourceType.virtual
+        )
+
+        self.delete_iRule = False
+
+    @staticmethod
+    def create_v4_iRule_content(tcp_options):
+        template = """when SERVER_INIT {
+            scan [getfield [IP::client_addr] "%%" 1] {%%d.%%d.%%d.%%d} a b c d
```
zongzw, line 39: Is it possible that customer or SEs inform us to change the content frequently. Is it necessary to put them into a separate configuration file, just like `f5-extended-profile.json`?

zongzw (approved):
lgtm. 
In our code, different features have some inter-relation at code level. Recommend we can have more different tests.

f5_openstack_agent/lbaasv2/drivers/bigip/irule.py
zhang-shengping: Yes, It may be possible. we need to confirm this with the SEs. If we allow an editable rule outside the code, it might be another feature?

zhang-shengping: > Is it possible that customer or SEs inform us to change the content frequently. Is it necessary to put them into a separate configuration file, just like `f5-extended-profile.json`?

Yes, It may be possible. we need to confirm this with the SEs. If we allow an editable rule outside the code, it might be another feature?

zongzw: > > Is it possible that customer or SEs inform us to change the content frequently. Is it necessary to put them into a separate configuration file, just like `f5-extended-profile.json`?
> 
> Yes, It may be possible. we need to confirm this with the SEs. If we allow an editable rule outside the code, it might be another feature?

It's ok to put it as another feature in the future. 
I just don't want to change the code every time according to requests from SE.

PR #1668, opened 2021-03-11
change travis images

zhang-shengping:
Instead, grab the docker image from the docker repository. we load docker image locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,14 @@ python:
 before_install:
     - git config --global user.email "OpenStack_TravisCI@f5.com"
     - git config --global user.name "Travis F5 Openstack"
-    - echo "$DOCKERHUB_TOKEN" | docker login -u "$DOCKERHUB_ID" --password-stdin
-    - docker pull f5devcentral/containthedocs:latest
+    - wget https://github.com/F5Networks/docker-images/raw/master/centos:6.tar.gz
+    - wget https://github.com/F5Networks/docker-images/raw/master/centos:7.tar.gz
+    - wget https://github.com/F5Networks/docker-images/raw/master/f5devcentral_containthedocs:latest.tar.gz
+    - wget https://github.com/F5Networks/docker-images/raw/master/ubuntu:trusty.tar.gz
+    - docker load -i centos:6.tar.gz
+    - docker load -i centos:7.tar.gz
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,8 +17,8 @@ python:
 before_install:
     - git config --global user.email "OpenStack_TravisCI@f5.com"
     - git config --global user.name "Travis F5 Openstack"
-    - echo "$DOCKERHUB_TOKEN" | docker login -u "$DOCKERHUB_ID" --password-stdin
-    - docker pull f5devcentral/containthedocs:latest
+    - git clone https://github.com/zhang-shengping/docker-images.git  
+    - docker load -i docker-images/f5_agent_travis_images.tar.gz
```
zhaoqin-github: I will recommend to use the docker  image name and tag to name the tarballs, so that others can understand which docker images are required easily.

.travis.yml
```diff
@@ -17,8 +17,11 @@ python:
 before_install:
     - git config --global user.email "OpenStack_TravisCI@f5.com"
     - git config --global user.name "Travis F5 Openstack"
-    - echo "$DOCKERHUB_TOKEN" | docker login -u "$DOCKERHUB_ID" --password-stdin
-    - docker pull f5devcentral/containthedocs:latest
+    - git clone https://github.com/F5Networks/docker-images.git
```
zhaoqin-github: Do not recommend to clone the whole repo, because it will fetch all branches, tags and files. If you add new image which are not required by this CI job, it will waste your network bandwidth.

Recommend to download your required image via raw file link, eg. https://github.com/F5Networks/docker-images/raw/master/ubuntu:xenial.tar.gz

zhaoqin-github (changes requested):

.travis.yml
zhang-shengping: No problem, I will change it part. Thanks for your review.

zhang-shengping: > Do not recommend to clone the whole repo, because it will fetch all branches, tags and files. If you add new image which are not required by this CI job, it will waste your network bandwidth.
> 
> Recommend to download your required image via raw file link, eg. https://github.com/F5Networks/docker-images/raw/master/ubuntu:xenial.tar.gz

No problem, I will change it part. Thanks for your review.

zhaoqin-github (approved):

PR #1688, opened 2021-04-22
Fix rds cache polluted problem

zhang-shengping:
This patch will only allow users to use their own nonshared
network, otherwise it will raise ProjectIDException and nothing
is created on bigip.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
@@ -287,9 +287,12 @@ class NetworkServiceBuilder(object):
             network['route_domain_id'] = 0
             return
 
-        LOG.debug("Assign route domain get from cache %s" % network)
+        LOG.debug("Assign route domain get from cache %s. "
+                  "The network is %s." % (self.rds_cache, network))
+
         try:
-            route_domain_id = self.get_route_domain_from_cache(network)
+            route_domain_id = self.get_route_domain_from_cache(
+                tenant_id, network)
             network['route_domain_id'] = route_domain_id
             return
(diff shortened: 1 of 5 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
```diff
@@ -287,9 +287,12 @@ def assign_route_domain(self, tenant_id, network, subnet):
             network['route_domain_id'] = 0
             return
 
-        LOG.debug("Assign route domain get from cache %s" % network)
+        LOG.debug("\nAssign route domain get from cache %s."
+                  "\nThe network is %s." % (self.rds_cache, network))
```
zongzw: We should split log lines into 2, instead of add'\n'. It's not benefit for log search.

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -239,6 +241,24 @@ def _pre_create(self, service):
...
+        lb_net_id = loadbalancer['network_id']
+        network = self.driver.service_adapter.get_network_from_service(
+            service, lb_net_id)
+
+        if not self.driver.network_builder.l2_service.is_common_network(
+                network):
+            net_project_id = network["project_id"]
+            if tenant_id != net_project_id:
+                raise f5_ex.ProjectIDException(
+                    "\nThe tenant project id is %s."
```
zongzw: We should split log lines into 2, instead of add'\n'. It's not benefit for log search.

zongzw (changes requested):

zongzw (approved):
lgtm.

PR #1689, opened 2021-04-22
Fix: project named route domain

zhang-shengping:
Problem:

the project named route domain will be created whenever the
max_namespaces_per_tenant is equal or large than one.

Resolution:

the project named route domain will only be created when the
max_namespaces_per_tenant is configured as 1.

the project named with auxiliary route domain will only be created
when the max_namespaces_per_tenant is configured large than 1.


zongzw (approved):
lgtm.

PR #1710, opened 2021-05-14
Catch all exception for route domain creation.

zongzw:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
@@ -399,7 +399,7 @@ class NetworkServiceBuilder(object):
                 LOG.debug("Allocated route domain %s for tenant %s"
                           % (rd_id, tenant_id))
                 break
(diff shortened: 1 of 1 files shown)


zhaoqin-github (changes requested):
First, should we ignore all route domain exception? Or we should only ignore route domain already exists exception?

Second, what is the root cause of route domain already exists exception? Does the code determine a wrong route domain id to create? Or it is caused by dirty data? Or we really need to ignore this duplicated route domain error?

zhaoqin-github (approved):

zongzw:
> First, should we ignore all route domain exception? Or we should only ignore route domain already exists exception?
> 
> Second, what is the root cause of route domain already exists exception? Does the code determine a wrong route domain id to create? Or it is caused by dirty data? Or we really need to ignore this duplicated route domain error?

The exceptions raised from `while` code are not only `f5_ex.RouteDomainCreationException`, but also `iControlUnexpectedHTTPError`(this time), so we use `Exception` to catch all kinds of them.

PR #1714, opened 2021-05-17
Upgrade eventlet to version 0.31.0 for Dependabot vulnerable.

zongzw:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


zhaoqin-github (changes requested):
If this requirement is higher than OpenStack pike requirement, I am not sure if it will have any negative impact. Maybe we need to have some test and also need to check with the users.

zhaoqin-github (approved):
It seems that we does not generate this prerequisite. Withdraw my previous input.

PR #1715, opened 2021-05-17
Upgrade eventlet to version 0.31.0 for Dependabot vulnerable.

zongzw:
(cherry picked from commit 43c03d3f4f320a5f951b27e0f4e7a696deb0cbdf)

@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/requirements.functest.txt
+++ b/requirements.functest.txt
@@ -7,7 +7,7 @@ python-coveralls==2.9.1
 mock==2.0.0
 coverage==4.2
 pytest==3.0.1
-eventlet===0.18.4
+eventlet>=0.31.0
 oslo.log===3.3.0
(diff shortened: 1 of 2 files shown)


zhaoqin-github (commented):
If this requirement is higher than OpenStack pike requirement, I am not sure if it will have any negative impact. Maybe we need to have some test and also need to check with the users.

zhaoqin-github (changes requested):
If this requirement is higher than OpenStack pike requirement, I am not sure if it will have any negative impact. Maybe we need to have some test and also need to check with the users.

zhaoqin-github (approved):
It seems that we does not generate this prerequisite. Withdraw my previous input.

PR #1716, opened 2021-05-17
Fix the issue of deleting nodes and snat in bigip ha mode.

zongzw:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -180,6 +180,7 @@ def delete(self, resource, service, **kwargs):
         payload = kwargs.get("payload",
                              self._create_payload(resource, service))
         LOG.debug("%s payload is %s", self._resource, payload)
+        self.driver.annotate_service_members(service)
```
Niklaus-xie: would need to double check if this affects other resources' deletion process, since this is a common place(from a base class)

jikui (approved):

PR #1743, opened 2021-06-03
Avoid patching profile failure, if profile is not created

zhaoqin-github:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
@@ -909,17 +909,42 @@ class ListenerManager(ResourceManager):
                 # properties defined in json file.
                 old_profile = {}
                 profile = {}
+                whole_profile = self.extended_profiles.get(profile_type, {})
                 customize = self.profile_map[profile_type].get("customize")
                 if callable(customize):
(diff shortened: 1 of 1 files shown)


zongzw (approved):

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -964,6 +992,33 @@ def _attach_profile(self, bigip, vs, profile):
...
+        }
+        v = self.resource_helper.load(bigip, name=vs['name'],
+                                      partition=vs['partition'])
+        profiles = v.profiles_s.get_collection()
+        for p in profiles:
+            if p.fullPath == loc:
+                attached = True
+                break
+            if p.fullPath == "/Common/http":
+                payload['profiles'].append(loc)
```
zhang-shengping, line 1013: What if loc is also an HTTP kind profile, of which the profiles will be effective? `/Common/http` or customized HTTP profile?  is it dependent on the sequence of the profiles in the python List?

Don't exactly understand the situation too much, Is there will be duplicated kinds of profiles in the payload['profiles']? What will happened if there are duplicated kinds of profiles.

PR #1746, opened 2021-06-15
Fix: if member is existed, it will cause HTTP 409 ERROR

zhang-shengping:

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
@@ -27,6 +27,7 @@ from f5_openstack_agent.lbaasv2.drivers.bigip import resource_helper
 from f5_openstack_agent.lbaasv2.drivers.bigip import tenants
 from f5_openstack_agent.lbaasv2.drivers.bigip.utils import serialized
 from f5_openstack_agent.lbaasv2.drivers.bigip import virtual_address
+from icontrol.exceptions import iControlUnexpectedHTTPError
 
 from oslo_log import helpers as log_helpers
 from oslo_log import log as logging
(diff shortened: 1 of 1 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -1154,11 +1155,20 @@ def _create_single(self, resource, service, **kwargs):
                 name=urllib.quote(pool_payload['name']),
                 partition=pool_payload['partition']
             )
-            pool_resource.members_s.members.create(**payload)
-            self._shrink_payload(pool_payload,
-                                 keys_to_keep=['partition',
-                                               'name', 'loadBalancingMode'])
-            self._pool_mgr._update(bigip, pool_payload, None, None, service)
+
+            try:
```
zhang-shengping, line 1159: I think it is better to eliminate this specific kind of Exception inside `MemberManager` rather than the upper level because we only try to catch the HTTP 409 exception caused by a single member create action?

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -1154,11 +1155,20 @@ def _create_single(self, resource, service, **kwargs):
...
+                pool_resource.members_s.members.create(**payload)
```
zhaoqin-github, line 1160: Do we only need to try/catch this line?

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -1154,11 +1155,20 @@ def _create_single(self, resource, service, **kwargs):
...
+                self._shrink_payload(pool_payload,
+                                     keys_to_keep=[
+                                         'partition',
+                                         'name', 'loadBalancingMode'])
+                self._pool_mgr._update(
+                    bigip, pool_payload, None, None, service)
+            except iControlUnexpectedHTTPError as ex:
+                if ex.response.status_code == 409:
+                    LOG.warning("The pool member %s is existed", payload)
```
zhaoqin-github: "is existed" --> exists

zhaoqin-github (changes requested):

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
zhaoqin-github: Sorry, I did not get your meaning......

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
zhang-shengping: Yes, it is not necessary to continue, if we get HTTP 409.

zhaoqin-github (approved):

Niklaus-xie (commented):
seems you need to re-submit the code in order to trigger travis.

PR #1749, opened 2021-06-23
udp/sip/diameter hm type

Niklaus-xie:


zhang-shengping (commented):
I do not know if SIP and Diameter health monitor are necessary? 
----> that is why I also submitted this PR. https://github.com/F5Networks/f5-openstack-agent/pull/1750 .
Anyway, since it is in a if/else, so adding sip+diameter does not harm, and also makes it more complete.

olvandeng:
> I do not know if SIP and Diameter health monitor are necessary?

We don't need to add SIP, or Diameter for now.

Niklaus-xie:
that is why I also submitted this PR. #1750 .
Anyway, since it is in a if/else, so adding sip+diameter does not harm, on the contrary, makes it more complete.

Afterall, we can choose between this PR and #1750 .


> > I do not know if SIP and Diameter health monitor are necessary?
> 
> We don't need to add SIP, or Diameter for now.

zhaoqin-github (approved):

PR #1771, opened 2021-08-25
redirect feature using irules

Niklaus-xie:


zhaoqin-github (changes requested):
Please add jira id into commit message. Thanks!

PR #1779, opened 2021-09-10
[OPENSTACK-2282]Fix resources creation conflict issue.

zongzw:
This issue happens in multiple agents environment.
When there are 2 or more loadbalancers are created at the same time,
the 2 loadbalancer creation logic from different agents may run into
the situation that both of them want to create the resources(of the same name).

This issue is similiar with route domain creation conflict described in PR 1654.
But the solution is different.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/l2_service.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/l2_service.py
@@ -15,6 +15,7 @@
 #
 
 import random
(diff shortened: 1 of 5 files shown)


zhaoqin-github (approved):

zhaoqin-github (changes requested):
Please add your jira id into commit message. Thanks!

zhaoqin-github (approved):

PR #1783, opened 2021-09-16
[OPENSTACK-2282] Cleanup redundant neutron port of self ip and snat ip

zhaoqin-github:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/selfips.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/selfips.py
@@ -180,6 +180,16 @@ class BigipSelfIpManager(object):
                 host_passed=host_passed
             )
 
+            if port:
+                ports = self.driver.plugin_rpc.get_port_by_name(
+                    port_name=selfip_name)
+                # If the port created by me is not the first, that means
+                # another agent process creates a new port before me.
+                # Delete my port and use the first one.
+                if len(ports) > 0 and ports[0]['id'] != port['id']:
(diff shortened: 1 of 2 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/selfips.py
```diff
@@ -180,6 +180,16 @@ def _get_bigip_selfip_address(self, bigip, subnet, device_id):
                 host_passed=host_passed
             )
 
+            if port:
+                ports = self.driver.plugin_rpc.get_port_by_name(
+                    port_name=selfip_name)
+                # If the port created by me is not the first, that means
+                # another agent process creates a new port before me.
+                # Delete my port and use the first one.
+                if len(ports) > 0 and ports[0]['id'] != port['id']:
```
zongzw, line 189: Shall we use `try .. except .. ` to catch port delete error, because there may happen that 2 agents do extra port deletions concurrently.

zongzw (approved):

f5_openstack_agent/lbaasv2/drivers/bigip/selfips.py
zhaoqin-github: > Shall we use `try .. except .. ` to catch port delete error, because there may happen that 2 agents do extra port deletions concurrently.

The first port got by those agent processes are identical, and each agent process will delete the port create by itself. There will not be any conflict.

PR #1857, opened 2022-04-27
L3 member

zhang-shengping:


f5_openstack_agent/lbaasv2/drivers/bigip/route.py
```diff
@@ -0,0 +1,55 @@
...
+        self.route_manager = BigIPResourceHelper(ResourceType.route)
+
+    def create_route(self, bigip, payload):
+        if payload:
+            try:
+                self.route_manager.create(
+                    bigip, payload
+                )
+            except Exception as err:
+                if err.response.status_code == 409:
```
zhaoqin-github, line 26: If dirty data exists, why not overwrite it?

Ideally we should place those common code in a base class, instead of implement them every time.

f5_openstack_agent/lbaasv2/drivers/bigip/pool_service.py
```diff
@@ -54,6 +54,71 @@ def __init__(self, service_adapter):
         self.node_helper = resource_helper.BigIPResourceHelper(
             resource_helper.ResourceType.node)
 
+    def delete_gateway_pool(self, bigip, pool_name, partition='Common'):
```
zhaoqin-github, line 57: I see too much generic resource create/delete procedure. I think we do need to consolidate them into a generic super class.

PR #1877, opened 2022-06-28
OPENSTACK-2489: connection limit

zhang-shengping:
set connection limit ratio according to virtual servers of a vip


zhaoqin-github (approved):

zhaoqin-github (changes requested):
withdraw this changes

PR #1969, opened 2022-11-14
[OPENSTACK-2626]raise exception when bigip status is not both active

zh805:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


zhaoqin-github (changes requested):
1. Need a jira id
2. I see some code dose not input ```no_bigip_exception``` when calling ```get_config_bigips()``` or ```get_all_bigips()```. If this parameter is always required, you may change its default value to True.

zhaoqin-github (approved):

PR #2002, opened 2023-02-15
[OPENSTACK-2624] Deploy configuration to multiple devices in parallel

zhaoqin-github:
@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


f5_openstack_agent/lbaasv2/drivers/bigip/utils.py
```diff
@@ -53,16 +53,27 @@ def real_serialized(method):
...
             my_request_id = uuid.uuid4()
 
             service = None
             if len(args) > 0:
                 last_arg = args[-1]
                 if isinstance(last_arg, dict) and ('loadbalancer' in last_arg):
                     service = last_arg
+
             if 'service' in kwargs:
                 service = kwargs['service']
```
zhang-shengping, line 67: 'service' is never in kwargs, and this service variable seems not important at all.

image

we need to find device info in service=args[-1].

But when here use it, args changes again.

image

finally, we may put service as a keyword argument.

zhang-shengping (changes requested):

f5_openstack_agent/lbaasv2/drivers/bigip/utils.py
```diff
@@ -53,17 +53,26 @@ def real_serialized(method):
...
 
+            # Construct a queue for each device
+            if "device" in service:
+                device_id = service["device"]["id"]
+                if "device_id" not in service_queue_map:
```
zhang-shengping: this should be `if device_id not in service_queue_map`

zhang-shengping (approved):

PR #2035, opened 2023-04-27
[OPENSTACK-2766] use mgmt_ipv6 in some cases

Niklaus-xie:
when use_mgmt_ipv6 is set to True in agent conf
file, and the mgmt_ipv6 address is included in
inventory db records, try to use the ipv6
management address for this bigip device.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/monitor_manager.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/monitor_manager.py
@@ -95,6 +95,10 @@ class LbaasMonitorManager(periodic_task.PeriodicTasks):
 
         self.system_helper = SystemHelper()
 
+        self.use_mgmt_ipv6 = False
+        if self.conf:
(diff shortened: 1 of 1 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/monitor_manager.py
```diff
@@ -444,9 +448,15 @@ def update_member_status_task(self):
                             m['device_info']['password']
                         )
 
-                        # TODO(nik) add v6 later when the config item is added
+                        if self.use_mgmt_ipv6 and m['mgmt_ipv6']:
+                            hostname = m['mgmt_ipv6']
+                            if ":" in hostname and \
+                               not hostname.startswith("["):
+                                hostname = "[" + hostname + "]"
+                        else:
```
zhaoqin-github, line 453: if this bigip only has one ipv6 addr, we will have to use ipv6 addr, even if it is not preferred.

zhaoqin-github (changes requested):

zhaoqin-github (approved):

PR #2056, opened 2023-06-16
[OPENSTACK-2806] Instantiating iControlDriver object in ResourceManag…

zh805:
…er to avoid using the same iControlDriver object when creating lb resources.

@<reviewer_id>
Fixes #<issueid>
WIP #<issueid>
...


zhaoqin-github (changes requested):
Please hold

PR #2058, opened 2023-06-16
9.10 fix delete

zhang-shengping:


f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
```diff
@@ -835,10 +836,12 @@ def annotate_service_members(self, service):
                 self.network_builder._annotate_service_route_domains(service)
             except f5ex.InvalidNetworkType as exc:
                 LOG.warning(exc.message)
+            except HTTPError as err:
```
zhaoqin-github, line 839: Does this change improve anything?

f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
```diff
@@ -411,13 +410,16 @@ def create_rd_by_net(
...
                 LOG.info("create route domain: %s, %s on bigip: %s"
                          % (name, route_domain, bigip.hostname))
-        except Exception as ex:
-            if ex.response.status_code == 409:
+        except HTTPError as err:
+            if err.response.status_code == 409:
                 LOG.info("route domain %s already exists: %s, ignored.." % (
-                    route_domain, ex.message))
+                    route_domain, err.message))
+            elif err.response.status_code == 400:
```
zhaoqin-github, line 417: Does this change improve anything?

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
```diff
@@ -1641,7 +1648,13 @@ def update(self, old_pool, pool, service, **kwargs):
     @serialized('PoolManager.delete')
     @log_helpers.log_method_call
     def delete(self, pool, service, **kwargs):
-        self.driver.annotate_service_members(service)
+        try:
+            self.driver.annotate_service_members(service)
+        except HTTPError as err:
+            if err.response.status_code == 400:
```
zhaoqin-github, line 1654: Why do we need to ignore it?

zhaoqin-github (changes requested):

f5_openstack_agent/lbaasv2/drivers/bigip/icontrol_driver.py
zhang-shengping: It will throw the original HTTPError, rather cover up by f5ex.RouteDomainCreationException.
The caller will not tell what Exception happens exactly.

f5_openstack_agent/lbaasv2/drivers/bigip/network_service.py
zhang-shengping: if it is 400, we throw up the originall HTTPError, rather than f5_ex.RouteDomainCreationException.

f5_openstack_agent/lbaasv2/drivers/bigip/resource_manager.py
zhang-shengping: > Why do we need to ignore it?

Olvan delete the whole partition, and nothing is left. When you try to annotate the route domain for members, the partition is disappeared, and the method returns 400 HTTPError. It will break the pool deleting process, it cause the pool cannot be deleted in the Neutron DB at last.

PR #2063, opened 2023-07-10
[OPENSTACK-2828] use same req id

Niklaus-xie:
when creating/deleting ports called from agent.


zhaoqin-github (changes requested):
Please hold a while. After we make other functional changes done, we can verify this change individually.

PR #2082, opened 2023-08-14
OPENSTACK-2840:  rebuild lb tree

zhang-shengping:
the 'network_id' is not in the member dictionary, when we rebuild loadbalancer.

so we need to have pre-conditions for different situations.


zhaoqin-github (changes requested):
What is the LB status during rebuilding?

zh805 (dismissed):

zhaoqin-github (approved):

PR #2101, opened 2023-09-07
[OPENSTACK-2875] not interrupt lb deletion process

Niklaus-xie:
not interrupt lb deletion process due to
tenant_cleanup. Concurrent deletion of lbs
might try to delete resources at same time.
The old way might reference some outdated
items which have been cleaned by other agents.
Try use helper.delete() to delete instead.


zhaoqin-github (changes requested):
Need to ensure to cleanup the partition after all lbs are removed

zhaoqin-github (changes requested):
PR https://github.com/F5Networks/f5-openstack-agent/pull/2103 works. Please close this one.

PR #2102, opened 2023-09-07
OPENSTACK-2840: Do not overwrite existing http2 profile.

zhang-shengping:
http2 profile uses default parameters, so its icontrol payload body only has 'name' and 'partition' attributes. These two attrs are immutable, the bigip device does not allow to patch (overwrite) them.


zhaoqin-github (commented):
Recommend to write your commit message a little more concise, eg. "Do not overwrite existing http2 profile".

zhaoqin-github (approved):

PR #2103, opened 2023-09-07
[OPENSTACK-2875] new way to delete in _remove_tenant_replication_mode

Niklaus-xie:
The old deletion way referenced some outdated
items which might have been cleaned by other agents. Try use helper.delete() to delete instead.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/tenants.py
@@ -15,8 +15,6 @@
 #
 
 from oslo_log import log as logging
-from requests import HTTPError
-from time import sleep
 
 from f5_openstack_agent.lbaasv2.drivers.bigip import exceptions as f5ex
 from f5_openstack_agent.lbaasv2.drivers.bigip.network_helper import \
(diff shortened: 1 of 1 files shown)


Niklaus-xie:
difference with 2101 is this still raises exception

zhaoqin-github (commented):
In the new config model, snat is associated with every loadbalancer. I assume they should be removed along with loadbalancer and needn't to be cleanup at the last step. Vlan, selfip and route domain are still shared resources, so that they need to be cleanup along with folder. Maybe we can exclude snat resource from this cleanup step. I am not very clear in which step route domain is removed. Please help to make sure the folder cleanup step is exactly correct , and please correct me if I am wrong. Thanks!

Niklaus-xie:
keeping the snat-X has one benefit though: it still tries to clean any snat-XX which has been left by whatever reason.   both acceptable to me whether removing the snat-XXXX part here or not . Personally I prefer not strictly  raising exceptions at this part.

zhaoqin-github (changes requested):
I confirmed that route is missed by this cleanup function. Just adding ResourceType.route before ResourceType.selfip (line 125), you code can cleanup the partition.

zhaoqin-github (approved):

PR #2120, opened 2023-10-18
[OPENSTACK-2899] SIP VS

Niklaus-xie:
based on protocol type(SIP), CRUD
the vs with /Common/sip and
new persistence profile.


Niklaus-xie:
only to re-trigger traivs. closed.

PR #2167, opened 2024-01-11
[OPENSTACK-2966] add purge in agent

Niklaus-xie:
Without deleting db data, this tries to purge
lb data on bigip device based on lb data from DB.

### DIFF
--- a/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager_lite.py
+++ b/f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager_lite.py
@@ -1763,3 +1763,99 @@ class LbaasAgentManager(periodic_task.PeriodicTasks):  # b --> B
                     "Fail to update status of loadbalancer %s "
                     "Exception: %s", lb_id, ex.message
                 )
+
+    @serialized('purge_loadbalancer')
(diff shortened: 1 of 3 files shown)


f5_openstack_agent/lbaasv2/drivers/bigip/agent_manager_lite.py
```diff
@@ -1652,3 +1652,92 @@ def rebuild_loadbalancer(self, context, loadbalancer, service):
...
+                mn_id = mn['id']
+                service["healthmonitor"] = mn
+
+                mgr = resource_manager.MonitorManager(
+                    self.lbdriver, type=mn['type'])
+                mgr.delete(mn, service)
+                LOG.debug("Finish to purge monitor %s", mn_id)
+            LOG.debug("Finish to purge all monitors in the lb")
+
+            for mem in members:
```
zhaoqin-github: I remember that all pool members should be removed along with the pool, after invoke a deleting pool request, so that we might not need to purge those pool members separately one by one before purging the pool. Can you please try it in that way?

zhang-shengping (approved):

zhaoqin-github (approved):

